PR #22, opened 2011-11-24
Various code fixes, a "new" environmental variables script and timestamp updating!

Jookia:
It's 7AM, I really don't want to explain it all. It's only a small diff, so just go through it, please?


Jookia:
Fixed your kind criticismseses.

src/shared/usercore/code/UpdateThread_Old.cpp
```diff
@@ -118,6 +121,10 @@ void UpdateThreadOld::doRun()
 				timer += boost::posix_time::minutes(1);
 			}
 
+			#ifdef NIX
+				UTIL::LIN::updateXDGRuntimeStamps();
```
lodle, line 125: this should be a generic function with the file name passed as an arg

Jookia, line 125: I disagree as that means every time we add a new file to the desura runtime dir you'd have to update this single point. It's simpler and easier to simply update all files in $XDG_RUNTIME_DIRS/desura and not worry about the specifics.
Also, if we did that, we'd have to worry about bugs that are only noticable due to files being removed on some systems every 6 hours, and it could be because of the programmer forgetting to add the file to this, or not knowing.

lodle, line 125: I do agree with that but the library where the function is stored is meant to be generic bunch of functions.

Jookia, line 125: I agree that it'd be nice, but sometimes it's better to DoTheRightThing than it is to make it generic if it adds complexity that could cause extremely hard to find bugs that'd only show up after running the client for 6 hours (AKA leaving your computer on overnight) on some systems.
It's not hard to rip out the existing code and change it, it's a single function.

However, if you actually think of the practice behind it, look at it this way:  Adding every file in $XDG_RUNTIME_DIRS/desura to an array for them to be updated. What happens if the file only exists for 12 hours? What about files that may be deleted, or not exist, on some systems? We'd have to have code to add/remove things from the array.  The current method avoids this.

Jookia:
You know what, it's been long enough.

PR #35, opened 2012-01-21
Fixed pthreads linking and a crash.

Jookia:
Infact, they may be related. Who knows.

### DIFF
--- a/build_lin/common.gypi
+++ b/build_lin/common.gypi
@@ -64,6 +64,7 @@
 		],
 		'ldflags': [
 			'-Wl,-Bsymbolic-functions',
+			'-lpthread',
 		],
 		'include_dirs': [
 			'<(include_dir)',
--- a/src/static/util/code/UtilLinux.cpp
+++ b/src/static/util/code/UtilLinux.cpp
@@ -194,8 +194,6 @@ public:
 	gcWString path;
 };
 
-WorkingDir g_WorkingDir;
-
 namespace UTIL
 {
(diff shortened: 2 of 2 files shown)


src/static/util/code/UtilLinux.cpp
```diff
@@ -245,7 +243,9 @@ bool is64OS()
 
 std::wstring getAppPath(std::wstring extra)
 {
-	gcWString wresult(g_WorkingDir.path);
+	WorkingDir dir;
```
Jookia, line 246: This may have the side effect of returning a different directory each time as it finds the directory each time. The keyword 'static' in front would probably fix it though.

src/static/util/code/UtilLinux.cpp
```diff
@@ -194,8 +194,6 @@ class WorkingDir
 	gcWString path;
 };
 
-WorkingDir g_WorkingDir;
```
Jookia, line 197: This may of been created in one thread, but had another thread trying to access it and failing, causing a crash on access.

build_lin/common.gypi
```diff
@@ -64,6 +64,7 @@
 		],
 		'ldflags': [
 			'-Wl,-Bsymbolic-functions',
+			'-lpthread',
```
Jookia, line 67: The only actual build affected by this is the bootloader, but I couldn't figure out how to add the flag to only it. Stuffing it in the common file worked though.

PR #38, opened 2012-01-22
Command to install all deps

yurikoles:
Small change in README file and initial install-deps.sh


Jookia:
I see this as potentially going somewhere, but there's no way we can track everything that third_party will need, and the script lacks a shebang and sudo.

PR #40, opened 2012-01-22
Dependencies installation for Fedora

yurikoles:
Install-deps.sh now determines running distribution and runs corresponding installation command. For now support is limited to Debian-based and Fedora.

### DIFF
--- a/README
+++ b/README
@@ -60,34 +60,11 @@ from the folder or setup visual studio to use the desura-d.exe from the build_ou
 Build Desura Linux
 ===================
 
-Need to have gcc 4.5 installed and these deps:
- * git-core   
- * subversion
- * m4
- * build-essential
- * binutils
- * automake
- * autoconf
- * libtool
- * libgtk2.0-dev 
- * libnss3-dev 
- * libgconf2-dev 
- * libgnome-keyring-dev 
- * libdbus-glib-1-dev 
- * gperf 
- * bison 
- * libcups2-dev 
- * flex 
- * libjpeg-dev 
- * libasound2-dev 
- * libbz2-dev 
- * libxpm-dev 
- * libx11-dev 
- * libssl-dev 
- * scons
-
-To install all of this deps on Debian/Ubuntu just invoke command:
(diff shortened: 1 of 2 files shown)


yurikoles:
To test build we have to resolve problem with wxwidgets checkouting.

svn: warning: Error handling externals definition for 'wxWidgets/build/tools':
svn: warning: URL 'http://svn.wxwidgets.org/svn/wx/wxWidgets/tools/build' at revision 70437 doesn't exist

lopho:
but its still a first point of orientation, better than knowing nothing about dependencies at all.

Jookia:
While this is a good idea, I have a tingling feeling that Bash scripts aren't the way to do this, as it already looks like a horrid mess. Could this be rewritten in Python? Could you use a hash map for each package and it's distro's package? Could we allow gcc 4.6 and bigger?

Jookia:
Alas, we have install-deps.sh already so I may as well merge this.

Jookia:
But please reconsider using Bash for this.

yurikoles:
Python is also dependency :)

Jookia:
Python and Bash are both dependencies.

yurikoles:
Bash is not required, the script is not bash-specific. 
Show me supported distribution without shell.

Jookia:
I'm saying that that's a shell script, we already need Python, so it's not another dependency being pulled in.

yurikoles:
Ok. What if user has no python installed? This script will install python as scons' dependency. When manipulations with system the shell is a must . We can also integrate support for other Unices.

Or we can install python with shell, then use python to install another dependencies. But “Beings
are not to be multiplied beyond necessity.”

Jookia:
If the user has no python installed, they can't build Desura, as requires gyp, which requires Python.

This script will turn in to a mess if it's in a shell script.

yurikoles:
> This script will install python as scons' dependency

Jookia:
Yes, but this script is going to end up being a very long and complex file, and I think rewriting it in Python could stop that, or other measures. On IRC, Lodle suggested having a script for each distro.

PR #41, opened 2012-01-22
on ubuntu or debian better to use sudo to get root access simply

hero1900:


yurikoles:
In default debian install user is not in sudoers.

hero1900:
okey then its better to make seperate bash for each platform what about fedora and other rpm based package
fedora and open suse using yum but then need the name of the packages.

yurikoles:
allredy done here  https://github.com/lodle/Desurium/pull/40
but there is no suse install command, just detection

PR #42, opened 2012-01-23
Use libnotify on Linux

Jookia:
Tada! This means you should see popups on Linux. testpopup works.


Jookia:
In the interest of public testing, I'm merging this in.

Issue #44, opened 2012-01-24
cURL version and Debian/Ubuntu


boskee:
cURL needs to be updated to at least 7.21.5 as 7.21.4 won't compile on Debian/Ubuntu because of no SSLv2 support

Jookia:
Apparently cURL in Desura is being patched to fix a multithreading DNS resolution problem. Contacting upstream and seeing if they have the fix would allow distros to use their own cURL, if the problem is fixed. I'll see what I can do there, but in the mean time it'd help if somebody (maybe even you) updated to .5.

boskee:
Sure. I've patched cURL to 7.21.5 and c-ares to 1.7.5. Nothing seems to be broken after the upgrade.

yurikoles:
Now curl won't build on Ubuntu. Here is solution.
http://curl.haxx.se/mail/archive-2011-12/0014.html

Jookia:
I don't understand that solution?

yurikoles:
Just read previous posts in that mail list thread. I got the same problem on Ubuntu.

Jookia:
Yes, but what do I do to fix it?

yurikoles:
Wee need to apply changes made by [that commit](https://github.com/bagder/curl/commit/2b0e09b0f98).

Jookia:
Have you tested to see if applying the changes will work?

Jookia:
Could you test this with the new build system?

ssokolow:
Possibly better. A fresh clone, having followed the instructions on Ubuntu 11.10 (Oneiric), now dies with this message:

```
Linking CXX executable mfc_upload_cli
/usr/bin/ld.bfd.real: ../../../external/curl/lib/libcurl.a(asyn-ares.o): undefined reference to symbol 'ares_process_fd'
/usr/bin/ld.bfd.real: note: 'ares_process_fd' is defined in DSO /usr/lib/gcc/x86_64-linux-gnu/4.6.1/../../../../lib/libcares.so so try adding it to the linker command line
/usr/lib/gcc/x86_64-linux-gnu/4.6.1/../../../../lib/libcares.so: could not read symbols: Invalid operation
collect2: ld returned 1 exit status
```

Jookia:
Okay, if you ad -lcares to somewhere in the cmake config? (Just go to the build dir and type 'ccmake' or 'cmake-gui .') Preferably the CXXFlags.

ssokolow:
[...]

PR #47, opened 2012-01-28
cURL 7.21.5 & c-ares 1.7.5

boskee:
Updated cURL to version 7.21.5 and c-ares to 1.7.5. Fixes Issue #44: cURL needs to be updated to at least 7.21.5 as 7.21.4 won't compile on Debian/Ubuntu because of no SSLv2 support


Jookia:
Does this include the patches?

Jookia:
Ah, it seems Lodle didn't separate the patches. This is a bit of an issue. Could you do a diff of the original third_party curl and c-are source code with the official versions they're derived from, to find the files Lodle edited of the old version?

boskee:
I did a diff on the original cURL 7.21.4 and original cURL 7.21.5 and patched third_party cURL with those changes. There was no rejects. Also, c-ares is pulled from its repo with third_party/SVNCheckout.sh script so there's nothing to patch.

Jookia:
So the updated curl has the same changes as ye olde curl?
Also, could you get a diff of the changes purely for curisoity's sake?

boskee:
Looks like my recent commit was included in this pull request. How can I exclude iand send it separately?

As for the diff - which one are you interested in? third_parthy/curl <> official 7.21.4 or official 7.21.4 <> official 7.21.5?

Jookia:
You'd have to branch it. I don't know.

I want third_party/curl_7.21.4 -> official 7.21.4

boskee:
Here you go http://pastebin.com/kfCsDURw

Jookia:
Sorry, but I'm going to have to close this on account of it being a better idea to just checkout curl from SVN and patching it. Importing the entire source tree isn't a good idea for every update.

boskee:
I agree, but right now curl is not in SVNCheckout scripts and there are no patches included. It's quite important to address, as right now Desurium fails to compile on Debian/Ubuntu because of that.

Jookia:
If you'd like to make that happen it'd be great. It's just build scripts being fixed.

PR #56, opened 2012-01-31
Fixed cef compile issue

LarryRotgut:
Warnings aren't treated as errors anymore when compiling cef


Jookia:
Other errors prevent me from compiling cef, so I'll just assume this works.

PR #71, opened 2012-02-29
updated readme and install-deps

yurikoles:
Removed dependecies list from readme, now it's refer to install-deps.sh, added build.sh

### DIFF
--- a/README
+++ b/README
@@ -5,7 +5,6 @@ Desura is a gaming client that allows a user to one click download and install g
 See here for more information: http://www.moddb.com/groups/desura/forum/thread/open-sourcing-desura
 
 
-
 Build Desura Windows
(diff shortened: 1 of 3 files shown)


yurikoles:
I think commands that user need to invoke before build except cloning git should be made ​​outside README

yurikoles:
I rechecked, last call ./build.sh is in directory build_lin, not in root of source tree.

Jookia:
Ah. This patch is great, exactly what I've been looking for. Though I don't see a justification for the Git cloning, seeing as the person may of obtained the source outside of Git, or have already used Git to get the source.

Jookia:
What I'm saying is, could you please remove the Git part?

PR #77, opened 2012-03-04
fix install-deps.sh for arch linux

matthiaskrgr:
Tested with the package "scons":
Before, I got "error: no targets specified (use -h for help)" after entering the password regardless "scons" being installed or not.
The new implementation removes the DEPS var (for arch) and processes the dependencies with "pacman -S --needed".

### DIFF
--- a/README
+++ b/README
@@ -125,4 +125,4 @@ cd build_lin
 Run Linux
 ===================
 
-If you built the debug build (default) you should be able to run ./test.sh in the build_lin folder to teset or ./desura in the build_out/debug_lin folder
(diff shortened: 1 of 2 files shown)


Jookia:
I'm a bit wary of this. Will this conflict if the user has gcc-multilib installed?

matthiaskrgr:
Indeed, my "fix" would suggest to uninstall gcc-multilib in this case (and install gcc).
Trying to get a fix done... :)

Jookia:
Thank you!

PR #78, opened 2012-03-06
Quote $DISTRIBUTION in install.sh for grep

jcnix:
Some Distros have a space in them, specifically Red Hat.
On a Red Hat distro install.sh will output the following
error when checking for Unknown distro:
grep: Hat: No such file or directory


Jookia:
Thank you!

PR #79, opened 2012-03-06
Add libstdc++-static to Red Hat/Fedora deps

jcnix:
bootstrap_lin passes -static-libstdc++ to g++ so it uses the static
libstdc++ instead of a shared library


Jookia:
Cheers!

Issue #61, opened 2012-02-03
Game status icons lack explanatory tooltips


ssokolow:
Yesterday, I had two games in my Linux Desura "Play" tab which were showing status icons with no clear means (beyond clicking them) for identifying what they indicate. (A pair of arrows in a circle and a lightning bolt)

As it's the most intuitive solution I'm aware of, given the current form integrated help in the Desura client takes (tooltips for some icons and a link to the forums), I propose that icons in the Play list should join toolbar icons in having explanatory tooltips.

Jookia:
I'd actually like to know about the icons too. I think the lightning means it's a demo. I'm not sure.

ssokolow:
I just tried clicking on them and the circle arrows signify the availability of an upgrade. (Why they didn't use a single upward-pointing straight arrow like Linux and proprietary app update managers use in their iconography rather than the icon browsers taught us to understand as "reload", I really want to know.)

As for the lightning bolt, clicking it did nothing but if your guess is correct, it's broken. The game with the lightning is World of Goo and I've had the full version installed from a Humble Bundle key since the first hour I had Desura installed.

However, World of Goo _is_ currently in a new Humble Bundle and I don't remember seeing a demo in my "Version" submenu before, so it might be that it's _supposed_ to mean "you're running a demo" but, due to a bug, actually means that a demo exists.

Protektor-Desura:
The circle arrows mean an update is available.
The lightening bolt means you are running a demo and it recommends you upgrade to the full version since you own the full version. The tooltip could be "Upgrade to your full version"

ssokolow:
Ahh, so it IS broken then.

I'm not the only person who's reported that, around the time the Humble Bundle for Android came out, World of Goo dropped from our public profile games lists and the lightning bolt appeared despite the full version being installed.

What's the proper place to report an issue of that sort. Would it be a server bug? Accidentally-messed up database records? I know it's not purely client-side because you can see World of Goo missing from my user profile if you visit Desura.com in Firefox.

Protektor-Desura:
The World of Goo must have freaked out when I updated it with a new version the other day. If the game should be in your account and is not then drop support an email so it can be forwarded to the right people to handle it.

ssokolow:
As far as I can tell from the contents of `html/playlist.html` in the default theme, there actually _should_ be tooltips, even in the release version.

[...]

PR #84, opened 2012-03-09
Fix for Desurium bug #61

Telemaniaka:
It fixes the tooltips on the "Update", "Demo", "Preorder" icons, and the Fav stars.

It does that by adding the jQuery.tipsy library to the page. Which is also used on the Settings page to display the tooltips there.


Jookia:
Thanks man!

PR #85, opened 2012-03-10
Run jpegoptim, optipng, advpng and pngcrush on 145 images.

matthiaskrgr:
248965 bytes -> 99963 bytes (-149002 bytes)
243 kb -> 97 kb (-145 kb)


matthiaskrgr:
The script I used can be found here: http://pastebin.com/NkYdwLtk

Jookia:
Sorry for the wasted effort man, but this only saves less than a half of a floppy disk, means we'd have to do the same for all future images to avoid having to do this again, and corrupts files. So I'm not going to merge it.

Keep up the good work, though!

PR #86, opened 2012-03-10
Translations for: 'Remove from account' is now 'Remove from list'; Added...

a-mroz:
... some translations which seemed to be missing


Jookia:
Thank you very much!

PR #101, opened 2012-03-25
Modified install-deps for ArchLinux, made the script less complicated

WhyNotHugo:
For ArchLinux, install-deps.sh made a list of dependencies, for queried which one where installed, then built a second list of packages, and called pacman with that list as an argument.

pacman can handle this itself, just using the "--needed" flag; it'll install required packages, and leave those present as they are.

This is just a simple change that removes unnecesary code, and it's quite cleaner _not_ to parse the output of pacman; the format can change at any time.

### DIFF
--- a/install-deps.sh
+++ b/install-deps.sh
@@ -11,14 +11,7 @@ if [ -f /etc/debian_version ]; then
 elif [ -f /etc/redhat-release ]; then
 	yum install git subversion m4 autoconf gcc-c++ libstdc++-static glibc-devel binutils autoconf libtool gtk2-devel nss-devel GConf2-devel libgnome-keyring-devel dbus-glib-devel gperf bison cups-devel flex libjpeg-turbo-devel alsa-lib-devel bzip2-devel libXpm-devel libX11-devel openssl-devel libnotify-devel scons xdg-user-dirs
 elif [ -f /etc/arch-release ]; then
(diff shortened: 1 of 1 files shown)


matthiaskrgr:
Mmmh,  when I run your configuration locally,  (sudo _line you add_), I get
...
resolving dependencies...
looking for inter-conflicts...
:: binutils and binutils-multilib are in conflict. Remove binutils-multilib? [y/N]

where my (current) implementation (running sudo ./install-deps.sh) says
"Dependencies already installed."

matthiaskrgr:
(note that "binutils-multilib" provides "binutils")

WhyNotHugo:
You're right, this does happen if you have another package "providing" one of the required ones. This sounds more like a bug in pacman though (it tries to install something "unneeded" when you use the "--needed" argument.) 

The issue in my case, was that I have an AUR helper aliased to "pacman", so that got called instead, and parsing the output did not work too well (though it would work fine in clean setup).  Perhaps I'll think of a cleaner way to do this, or this details will get fixed in future in pacman though.

In any case, if you select "N" as a response to that, it does work fine, right?

matthiaskrgr:
resolving dependencies...
looking for inter-conflicts...
:: binutils and binutils-multilib are in conflict. Remove binutils-multilib? [y/N] n
error: unresolvable package conflicts detected
error: failed to prepare transaction (conflicting dependencies)
:: binutils and binutils-multilib are in conflict

Jookia:
matthias's problem is why the workaround is there.

Jookia:
I don't think this is going to work, due to matthias' problem, which is already being worked around. If it's a bug in pacman, take it up with the pacman authors.

PR #128, opened 2012-03-29
implement full debug builds in cmake for linux and some #130 stuff

karolherbst:


Jookia:
I find this redundant as most external libs aren't debug since they're installed by package managers.

karolherbst:
but on windows we must to have an option in cmake to build external debug libraries. At least for boost and wxWidgets. So I think a better way to do this would be the usage of an option like: DEBUG_EXTERNEL.

As for gentoo there are a debug option for most of the libraries to build both: release and debug libs.

Jookia:
Hmm.. DEBUG and DEBUG_EXTERNAL , two separate objects. I really like the idea.

Jookia:
Reviewing the patch, you removed the ability to use a custom branding, and changed 'Desura' to 'Desurium'. :l

karolherbst:
you're right, didn't thought about that.

PR #142, opened 2012-04-05
start updating PKGBUILD, unbreak build for gcc 4.7

matthiaskrgr:
This MR starts updating the PKGBUILD (wasn't able to fully test yet due to failure of getceflibs.sh) and fixes the build for compilation with gcc 4.7 (thanks to Karol for hints).


Jookia:
How tested is this?

matthiaskrgr:
Fixed compilation with gcc 4.7 and some other stuff which prevented makepkg from running successfully.
A package is now created but to run the binary some more tweaks are needed.

getceflibs.sh
```diff
@@ -20,11 +20,14 @@ else
...
-mkdir $COPYPATH
+if [ ! -d "${COPYPATH}" ] ; then
+	mkdir $COPYPATH
+fi
+echo "Downloading libs..."
 wget $URL -O desura.mcf
 export LD_LIBRARY_PATH="`pwd`/install/lib"
-install/bin/mcf_extract desura.mcf tmp_desura
+echo "Extracting libs..."
+./build/src/tools/mcf_extract/mcf_extract desura.mcf tmp_desura
```
Jookia: how will this fine the libs it uses as above it sets the LD_LIBRARY_PATH to `pwd`/install/lib

matthiaskrgr: I had to change this because "install/bin/mcf_extract" didn't exist.
Perhaps we should try to get rid of the pwd somehow and use "./bla" ?

matthiaskrgr: Also I'm not really convinced that I introduced the issue. I just fixed the path to mcf_extract which wasn't found before.

Jookia: You need to run 'make install'.

matthiaskrgr:
What is archived by now:
Desura build using gcc 4.7 (thx to karol!)
getceflibs.sh doesn't fail because it cannot find some binaries
overhaul of the PKGBUILD (also add desktop files)
add a execution script to /usr/bin which runs desura from /opt

issues: when I access getceflibs.sh using the pkgbuild, I get
Copying libs to destinations...
cp: cannot stat ‘/lib/libcef_desura.so’: No such file or directory
(maybe because of this) desura crashes very easily

Jookia:
I pulled this in manually.

PR #146, opened 2012-04-09
updated ebuild for gentoo

karolherbst:
whats not working:

download libcef (you have to build it ;) )


karolherbst:
I do something bad

PR #159, opened 2012-04-16
apt: use apt-get instead off aptitude and added cmake to deps

karolherbst:


LiamDawe:
What difference does it make using apt-get instead of aptitude? (Just curious)

ssokolow:
@liamdawe: apt-get is installed on every Debian-based distro I'm aware of while aptitude is excluded from the default installs of Ubuntu-family distros. (You have to explicitly use `apt-get install aptitude` on Ubuntu and friends)

PR #170, opened 2012-04-20
Improve cef

karolherbst:
we should download chromium sources instead of gclient checkout.


Jookia:
Does this work?

karolherbst:
for me, yes

PR #175, opened 2012-04-21
Pkgbuild

matthiaskrgr:


matthiaskrgr:
Mmh, this doesn't seem to work as well as intended

Jookia:
Does it work now?

matthiaskrgr:
I didn't test yet, but it should work now, unless there are problems with --depth 1

Jookia:
Those sound like famous last words. Ah well.

PR #186, opened 2012-04-24
lets give a try

### LINKED ISSUE #33: Upload & Download issue
This one is going to be a pain in the behind to find. Consitantly but with no rhyme or reason the Linux client will crash and exit when doing an upload. The Windows version doesn't seem to have this problem. Sometimes it happens immediately when you try and start the upload sometimes it take as long at 10 minutes. It always crashes though before the upload is finished unless it is a small file like under 10 megs. Most updates to games are this small but getting a game up under Linux it always crashes multiple times and I have to resume it.

Less frequent is the crash on download. I have I would say maybe 5% of the time the client will crash on a download both on Windows and Linux. 

I do way more uploading and downloading than most do with the client, so I see it all the time, but I am getting reports from developers and users of this issue.

karolherbst:
fixes #33

tested it while downloading 6 games (4GB) at once. No crash
thesiege(IRC:anagromataf) tested while downloading 50 games.


Jookia:
I uploaded 1GB of illegal music. Didn't crash.

PR #187, opened 2012-04-25
we need to link against pthread on ubuntu

karolherbst:


Jookia:
I'm a bit worried that this'll break on Windows.

karolherbst:
oh yes, you are right

PR #194, opened 2012-04-26
Ubuntu stuff

karolherbst:
some deps


karolherbst:
on ubuntu and mint libcrashuploader.so will not linked against libcares.so because it doesn't find any references in libcrashuploader.so itself. But because we build our CURL staticly (so we can't get confused with a system curl, which is built without ares support, but threaded) ld does not find the references.

Simple Solution: we pass cares as a link target after curl, so ld will find it

fixes #196

PR #197, opened 2012-04-27
install-deps.sh: arch: add yasm

matthiaskrgr:


Jookia:
We may also need to note that we need libxml2 and libevent.

PR #231, opened 2012-05-24
getCmdStdOut now reads from popen correctly

Smilex:
The problem was that fgets was used to read before, although it is not designed to read binary data, so I used fread instead.
This fix also made getOSString() print correctly


Smilex:
There are in fact some serious bugs in this pull request, so refrain from pulling it

PR #240, opened 2012-06-08
concerning issue #190

Smilex:
Can't promise that this fixes issue #190
But I did change the class ExeDir into a function called getExecuteDir(). I also added a Unit Test, which for me, works correctly.


Jookia:
Looks good. I'll check it out after I fix karolherbst's code.

karolherbst:
I have merge these commits into the branch "bug190". It would be nice if someone would test it there

karolherbst:
it is pulled into a branch already

PR #242, opened 2012-06-11
Update Debian related variables

makson96:
Few notes:
- in CPACK_DEBIAN_PACKAGE_MAINTAINER there is email address missing in "<>" marks.
- in CPACK_DEBIAN_PACKAGE_DEPENDS I needed to put some dev packages, because packages with standard libraries differ by names in different version of systems and I want to make it work on as many systems as possible (dev package will install standard library as a dependency).
- on Ubuntu Natty and Debian Squeeze and earlier versions there is no libwebp, so dependencies could not be satisfied.

### DIFF
--- a/cmake/modules/Package.cmake
+++ b/cmake/modules/Package.cmake
@@ -12,6 +12,7 @@ set(CPACK_SOURCE_GENERATOR "TBZ2;TGZ;ZIP")
 ###############################################################################
 
 set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Desura is a gaming client that allows users to one click download and install games and game modification.")
+set(CPACK_PACKAGE_HOMEPAGE "http://www.desura.com/")
 set(CPACK_PACKAGE_VERSION_MAJOR 0)
 set(CPACK_PACKAGE_VERSION_MINOR 8)
 set(CPACK_PACKAGE_VERSION_PATCH 0)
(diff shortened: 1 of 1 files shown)


makson96:
I have also noted, that depends in Debian control file are separated by ", " while in CPACK_DEBIAN_PACKAGE_DEPENDS there is no comma between them. For the moment I left it that way, because I am not sure if it is intentional.

karolherbst:
did you test creating a deb file with "make package" in the build directory?

karolherbst:
and libwebp is no longer a dependency of desurium, because of the same problem you figured out.

makson96:
I have missed libwebp removal. I will remove it from dependencies in next fix. :)

Also architecture "any" will not work. Need to do some condition to set is as a "amd64" or "i386" (maybe this will be helpful: http://www.mail-archive.com/cmake@cmake.org/msg12849.html).

Package manager will complain that there is no email for maintainer (do you want to put there yours or you want me to sign as a maintainer?)

Maybe I should set CPACK_DEBIAN_PACKAGE_VERSION to "${CPACK_PACKAGE_VERSION}"?

The package do not create any menu entry or link in /usr/bin. So desurium could only be run by invoking run.sh from installation dir.

After setting correct architecture package create itself and install correctly on Ubuntu 12.04.

karolherbst:
sounds nice.

We have a 64BIT Variable in cmake, so it is able to check via "if(64BIT)"

If you want you can maintain the deb package.

At the moment we don't know how we should set version number and stuff, so something like "0.0.1" would be fine at the moment.

Normally symlinks are created automatically. Look here: https://github.com/lodle/Desurium/blob/master/src/CMakeLists.txt#L131
At the moment it isn't a good idea to install desura in /usr

It would be better to configure the project with

```
cmake .. -DPACKAGE_TYPE="DEB" -DCMAKE_INSTALL_PREFIX="/usr/games/desura"
```

(see: https://github.com/lodle/Desurium/blob/master/README#L71)

makson96:
This is it. Package now builds correctly on my machine and install and work on fresh Ubuntu installation (on VirtualBox).

PR #244, opened 2012-06-12
Added ./ to sudo install-deps.sh

madrang:


ssokolow:
Serves me right for leaving that change in my local copy for weeks without submitting a pull request.

PR #252, opened 2012-06-24
Bug190

### LINKED ISSUE #190: No language loaded.
I don't know what causes this, but sometimes the language strings aren't loaded.

karolherbst:
this could fix #190

I can't test it, because it didn't occure from me, but it is building here

compile works for:
- karolherbst (gentoo, gcc-4.7.1, clang-3.1)
- anagromataf (gentoo, gcc-4.7.1)

### DIFF
--- a/src/include/util/UtilLinux.h
+++ b/src/include/util/UtilLinux.h
@@ -38,6 +38,20 @@ namespace LIN
 		BT_SCRIPT,	// Launch using xdg-open
 		BT_UNKNOWN,	// Error
 	};
+
+	/*! \brief Gets the path from which the executable was launched.
+	 *  \return Returns a std::string for the path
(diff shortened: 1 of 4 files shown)


Jookia:
Could you explain _why_ it would fix it?

karolherbst:
maybe I will change it to "could" because Smilex wasn't sure about this

TheSiege:
compiles and runs; though I can't reproduce #190; I can't say if it is fixed.
gentoo, gcc 4.7.1

Jookia:
I'm not convinced that this fixes it until we can reproduce the bug. What're the other benefits to this patch?

karolherbst:
I rebased this request on newest master

TheSiege:
compiles and runs on
ubuntu 12.04
fedora 17

PR #256, opened 2012-07-04
Cmake patch script improvements

karolherbst:
this is usefull if we want to modify some patch files.

It will create a backup of every used patch file and will revert projects with this backup if needed, so we don't have to remove the whole project anymore

reviewed by:
- matthiaskrgr


Jookia:
Great idea, looking through the code it looks fine but I'm wary of shell scripts, so I don't want to be the one to break the build by merging. You have my vote.

PR #257, opened 2012-07-04
some cmake improvements and README update

karolherbst:
works for:
- karolherbst (gentoo + ebuild, gcc-4.7.1)
- TheSiege (gentoo. gcc-4.7.1 / linux mint 13, gcc-4.6.3)


Jookia:
Why not just remove the SET_OWN_EXT_SRC flag, try and set the variables to the user's inputs by default and fall back on the predefined stuff? It'd allow individual sources and remove the need for all the error messages.

karolherbst:
yeah, I will think about this over night

TheSiege:
We could do that but then I strongly argue for being able to set the svn repos too

karolherbst:
I don't know how to handle hashes in this case. We could enforce specific versions of packages, but I don't like the idea, because there are maybe some reasons to use a newer or older version

TheSiege:
I'd say don't do hashes in this case. So one can experiment.

karolherbst:
yes, but I want to avoid if clauses

karolherbst:
but I have an idea

karolherbst:
I rebased this request on current master so we are able to merge this cleanly. Work seems to be done. Open for testing.

TheSiege:
works on:
gentoo gcc-4.7.1
linux mint 13 gcc-4.6.3

PR #262, opened 2012-07-07
Pkgbuild rm unused deps

matthiaskrgr:


karolherbst:
it should be tested in a clean arch install (with the core system only, smallest possible install, etc..)

PR #265, opened 2012-07-11
System wxWidgets

karolherbst:
these changes will allow us to use a system wide installed wxWidgets.

works for:
- karolherbst

PR #272, opened 2012-07-23
Allow to compile with system breakpad.

aszlig:
The patch for Google Breakpad only seems to be needed when building for Windows, so I guess we can safely allow to use breakpad from the system.

This commit depends on b262dff51214439f8713523f94a92b82ae602e30 from pull request #265, as it contains `$(REQUIRED_IF_OPTION)`, as we don't want to force non Windows users to actually have breakpad available (do we? :smile:).


aszlig:
Reopened with changed base branch.

PR #273, opened 2012-07-24
Allow to compile with system breakpad.

aszlig:
The patch for Google Breakpad only seems to be needed when building for Windows, so I guess we can safely allow to use breakpad from the system.

This commit depends on b262dff51214439f8713523f94a92b82ae602e30 from pull request #265, as it contains `$(REQUIRED_IF_OPTION)`, as we don't want to force non Windows users to actually have breakpad available (do we? :smile:).


karolherbst:
a few things about this request:
1. nearly no distribution has the ability to install breakpad system wide, so "${REQUIRED_IF_OPTION}" will break the build on nearly 99% of all systems.
2. breakpad itself doesn't have a proper make install target, see http://code.google.com/p/google-breakpad/issues/detail?id=219
3. you are depengin on the system_wx pull request because of the "${REQUIRED_IF_OPTION}"

I would say we should not require breakpad anyways, if it is found, then nice, but I don't want have a possibility to force it at the moment.

TheSiege:
maybe it should be a $REQUIRED_BREAKPAD option then because it is only on very few systems

Mailaender:
There are some packages of google-breakpad for OpenSUSE. It would be nice to have an option to have one included dependency fewer to get his into distributions.

hasufell:
Gentoo would include google-breakpad too, there is already a patch available for the install rules, but we want upstream to include a .pc file as well as you can see on the referenced issue.

lodle:
If the community doesnt care about crash dumps you shouldnt even need break pad in desura

karolherbst:
and what about the official desura client? But I like the idea to remove breakpad, because I don't want to upload every crash somewhere.

But it would be nice to have some crash information. But this should be possible via signal.h and some backtracing.

lodle:
should be easy to make it a conditional compile

PR #278, opened 2012-07-31
Improved on GetAppBuild() and GetAppBranch()

Smilex:
This is nothing major, but those two functions should now be prettier and faster, by removing all the unnecessary conversions done by it previously.


Jookia:
20 commits worth of nothing major?

karolherbst:
what the heck, lern to rebase and branch :p

PR #279, opened 2012-07-31
cmake: add chromium patch for bison-2.6 compatibility

### LINKED ISSUE #275: desurium won't compile with bison 2.6
I constantly get an compiling error on gentoo with cef.
https://gist.github.com/3173663

it seems to be a namespace error because adding on line 480 :: before the functinon xpathyyparse in
Desurium/build/chromium-prefix/src/chromium/src/third_party/WebKit/Source/WebCore/xml/XPathParser.cpp
fixes it for me

karolherbst:
fixes #275

open for testing

compiles for
- karolherbst (bison-2.6.0, bison-2.5.1)
- TheSiege (bison-2.6.0, biston-2.5.1)

### DIFF
--- a/cmake/modules/BuildCEF.cmake
+++ b/cmake/modules/BuildCEF.cmake
@@ -99,6 +99,14 @@ if(NOT WIN32)
 
   ExternalProject_Add_Step(
     cef
+    bison-2-6-patch
+    COMMAND ${CMAKE_SCRIPT_PATH}/patch.sh ${CMAKE_SOURCE_DIR}/cmake/patches/chromium-bison-2.6.patch
+    DEPENDERS patch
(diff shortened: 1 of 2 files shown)


TheSiege:
gentoo; gcc-4.7 compiles and runs fine for bison 2.6.0 and 2.5.1
also works on ubuntu 12.04 with bison 2.5
also works on fedora 17 with bison 2.5

Smilex:
Runs for me with bison 2.5.1

matthiaskrgr:
PKGBUILD compiles fine with bison 2.6.1-1 on arch.

Cheeseness:
I've compiled and tested that this runs on F16 with bison 2.5-1 :)

PR #281, opened 2012-08-01
Update of install-deps.sh for Fedora

Cheeseness:
anagromataf mentioned that his base F17 install didn't include the patch package, so that has been added to the script.


Cheeseness:
Whoops. Seems I somehow included the bison-2.6 branch accidentally.

PR #284, opened 2012-08-01
added perl-Digest-MD5 dependency for fedora 17

### LINKED ISSUE #280: patch needs to be included in install-deps for fedora and some perl md5 disgest stuff
patch needs to be included in install-deps for fedora
tested on clean fedora 17 install
Desurium/cmake/scripts/patch.sh" line 19: patch: command not found
also http://pastebin.com/Wv6RTmqX happens which I fixed with installing
perl-Digest-MD5

TheSiege:
needed to fix #280


karolherbst:
is this needed by cmakes MD5 hashing for external projects?

karolherbst:
I should read bug280 before.. okay

PR #292, opened 2012-08-14
Fix for #223: Locally-installed entries are forced to "Not Installed" on restart

### LINKED ISSUE #223: Locally-installed entries are forced to "Not Installed" on restart
When I switched from the latest non-GPL Desura (Linux version) to revision 3a317cf5201efb96469434b4f320971ae256c544, all of my locally-installed games were marked as "Not Installed".

In the process of trying to compare an identical "Installed" and "Not Installed" entry, I discovered that, while an "Installed" entry can be created (In my case, for `/usr/games/frogatto`) and will execute properly, it will reset to "Not Installed" when Desura is restarted. (cleanly or via killall and `SIGTERM` as I use when the non-GPL version won't exit)

I've also confirmed that the contents of `iteminfo_d.sqlite` for newly-created entries are identical to the old ones migrated over. (I deleted and re-created the one for `/usr/games/frogatto` and only the primary key changed)

r-ick:
This fixes #223 for me, plus a related bug I found in testing it where a 64-bit only build refuses to run local links to 64-bit binaries with an error about 32-bit compatibility.

I've gone for the solution of regenerating each link's BranchInfo on start, since enough data **is** succesfully saved to do that and I can't see anything to suggest a "proper" branchid range they should be given instead.

Edit: I've also cleaned up some #ifdefs that are basically identical to my own.

works for:
- lodle (Windows)
- TheSiege (Linux 64)

### DIFF
--- a/src/shared/usercore/code/ItemInfo.cpp
+++ b/src/shared/usercore/code/ItemInfo.cpp
@@ -177,9 +177,12 @@ void ItemInfo::saveDb(sqlite3x::sqlite3_connection* db)
 		cmd.bind(9, (long long int)m_iId.toInt64()); //internal id
 		cmd.executenonquery();
 
-		for (size_t x=0; x<m_vBranchList.size(); x++)
-		{
-			m_vBranchList[x]->saveDb(db);
+		// Don't bother saving TYPE_LINKs, they don't have a proper BranchID anyway.
+		if(getId().getType() != DesuraId::TYPE_LINK) {
+			for (size_t x=0; x<m_vBranchList.size(); x++)
+			{
+				m_vBranchList[x]->saveDb(db);
+			}
 		}
 
 		std::for_each(m_mBranchInstallInfo.begin(), m_mBranchInstallInfo.end(), [&db](std::pair<uint32, BranchInstallInfo*> p)
@@ -225,9 +228,12 @@ void ItemInfo::saveDbFull(sqlite3x::sqlite3_connection* db)
 
 	cmd.executenonquery();
 
-	for (size_t x=0; x<m_vBranchList.size(); x++)
-	{
-		m_vBranchList[x]->saveDbFull(db);
+	// Don't bother saving TYPE_LINKs, they don't have a proper BranchID anyway.
+	if(getId().getType() != DesuraId::TYPE_LINK) {
+		for (size_t x=0; x<m_vBranchList.size(); x++)
+		{
+			m_vBranchList[x]->saveDbFull(db);
+		}
 	}
 
 	std::for_each(m_mBranchInstallInfo.begin(), m_mBranchInstallInfo.end(), [&db](std::pair<uint32, BranchInstallInfo*> p)
--- a/src/shared/usercore/code/ItemInfo.cpp
+++ b/src/shared/usercore/code/ItemInfo.cpp
@@ -356,15 +356,7 @@ void ItemInfo::loadDb(sqlite3x::sqlite3_connection* db)
 
 		// A TYPE_LINK may or may not have branches stored on disk, but will need one at runtime.
 		if(getId().getType() == DesuraId::TYPE_LINK && m_vBranchList.size() == 0) {
-#ifdef WIN32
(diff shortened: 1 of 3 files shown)


src/shared/usercore/code/ItemInfo.cpp
```diff
@@ -347,6 +353,22 @@ void ItemInfo::loadDb(sqlite3x::sqlite3_connection* db)
 
 			m_vBranchList.push_back(bi);
 		}
+
+		// A TYPE_LINK may or may not have branches stored on disk, but will need one at runtime.
+		if(getId().getType() == DesuraId::TYPE_LINK && m_vBranchList.size() == 0) {
+#ifdef WIN32
+			BranchInfo* bi = new BranchInfo(MCFBranch::BranchFromInt(m_INBranch), m_iId, m_mBranchInstallInfo[100]);
+#else
+#ifdef NIX64
```
karolherbst: could you use here #elsif instead of another nested #ifdef clause?

Best would still the usage of a static #ifdef set variable, so this #ifdef mess is fully eliminated here.

lodle:
Good fix. Links where only tested on windows :P

src/shared/usercore/code/ItemInfo.cpp
r-ick: That would be a better idea, but rather than a static variable, can you think of any reason not to use BUILDID_PUBLIC from Common.h? It would help clean up at least one similar structure in ItemInfo::setLinkInfo() too.

karolherbst: this would be a good solution, yes use BUILDID_PUBLIC.

karolherbst:
looks good now.

I would say open for testing now! (so no commits please anymore here)

TheSiege:
so I tested it now and .... now manually adding stuff does not work anymore for me

r-ick:
In what way?

TheSiege:
in the way; that when I try to add something it does not show up after I click add

r-ick:
Are you deleting an existing link then immediately re-creating it with the same name and exe target without restarting? That's the only situation where I've seen that happen on my branch or master, and AFAICT it's a different bug that's nothing to do with me or #223 :)

Otherwise, I'm not sure what I've touched that could possibly stop links from appearing at all.

TheSiege:
ok; that seems to be right; now when I run the run.sh I still get the error with not installed

r-ick:
Is that with new links or just existing ones or both?

TheSiege:
I had no existing ones, so new ones

r-ick:
Hmm. Try running this in a terminal after creating some links then exiting Desurium:

``` bash
sqlite3 ~/.desura/iteminfo_d.sqlite "select * from iteminfo where statusflags & (1<<21); select * from branchinfo where internalid in (select internalid from iteminfo where statusflags & (1<<21)); select * from exe where itemid in (select internalid from iteminfo where statusflags & (1<<21)); select * from installinfo where itemid in (select internalid from iteminfo where statusflags & (1<<21))"
```

What is the output?

karolherbst:
what is the current status here?

r-ick:
Waiting for @TheSiege, or another tester, to post something that tells me why my patches are (aparrently) having no effect for him.

TheSiege:
```
 in (select internalid from iteminfo where statusflags & (1<<21))"
-175739448888131504|0|0|2129946|||egoboo||||||||||0|0
5271338174130421840|0|0|2129946|||ttd||||||||||0|0
-175739448888131504|120|Link|/usr/games/bin/egoboo|||-1
5271338174130421840|120|Link|/usr/games/bin/openttd|||-1
-175739448888131504|120|/usr/games/bin|/usr/games/bin/egoboo||0|0|0
5271338174130421840|120|/usr/games/bin|/usr/games/bin/openttd||0|0|0
```

TheSiege:
also; I can't reproduce my previous encounter anymore; it is fixed for me now

r-ick:
Fair enough then... and a good thing too, because there's nothing I can see wrong with that DB dump.

karolherbst:
@TheSiege
what is with this? "so I tested it now and .... now manually adding stuff does not work anymore for me"

PR #295, opened 2012-08-25
cmake: add Boost Include Dir to test targets

karolherbst:
this will fix compile errors, if no /usr/include/boost directory is existing on system. This happens for example on gentoos boost-1.50-r2 ebuild


TheSiege:
fixes issue compiles fine; please merge

matthiaskrgr:
compiles fine

PR #298, opened 2012-08-28
some minor improvements

karolherbst:
there is nothing special, but we should make all destructors virtual, because they aren't added by default and will cause mem leaks in subclasses

works for:
- karolherbst


karolherbst:
this should not break anything. I will merge this on Sunday, if nobody complains

PR #303, opened 2012-08-30
full FHS support

### LINKED ISSUE #296: Give an option to be FHS compliant
A few hours ago I chatted a little with an openSUSE (@Mailaender) and an Gentoo (@hasufell) package maintainer. the rpm based distro would like an option to be FHS (https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) compliant; while the Gentoo person demands it in order to be in the repositories. Essentially we have been FHS compliant one time but this: https://github.com/lodle/Desurium/commit/07fde3828d86c2ce81d29bea2984a48b42af43c1 commit broke it.

EDIT: karolherbst

we have to do this:
- create a system_path API
- let cmake configure a source file, so we have this absolute paths insource
- allow cmake to specify every needed path in an absolute way
- use the system_path API in the code

karolherbst:
this will implement full FHS support

fixes #296

works for:
- karolherbst (local, ebuild)
- TheSiege (local, ebuild)
-  Mailaender (rpmlint)


src/static/util/code/UtilOs.cpp
```diff
@@ -135,13 +135,6 @@ void setConfigBinaryValue(const std::string &regIndex, const char* blob, size_t
 #endif
 }
 
-std::wstring getDataPath(std::wstring extra)
-{
-	std::wstring dataDir = L""; // Use to be 'data'.
-	
-	return getCurrentDir(dataDir + extra);
-}
-
```
Jookia, line 144: You'd think this would break something?

build_out/run.sh
```diff
@@ -5,6 +5,9 @@ SOURCE="${BASH_SOURCE[0]}"
...
 
-export PATH="$DIR:$PATH"
-export LD_LIBRARY_PATH="$DIR/lib"
-desura $@
+if [[ "@RUNTIME_LIB_INSOURCE_DIR@" = /* ]]; then
+	LD_LIBRARY_PATH="@RUNTIME_LIB_INSOURCE_DIR@/" @RUNTIME_LIB_INSOURCE_DIR@/desura $@
+else
+	LD_LIBRARY_PATH="$DIR/@RUNTIME_LIB_INSOURCE_DIR@/" $DIR/@RUNTIME_LIB_INSOURCE_DIR@/desura $@
+fi
+
```
Jookia, line 13: The readme says not to use run.sh What's it still here for?

Edit: Oh. Shouldn't it be renamed then?

cmake/modules/SetInstallPaths.cmake
```diff
@@ -1,17 +1,48 @@
...
 set(CMAKE_INSTALL_PREFIX ${DEFAULT_INSTALL_DIR}
     CACHE STRING "Desura Install Prefix")
+set(BINDIR ""
+    CACHE STRING "Desura Bin Install Dir")
+set(RUNTIME_LIBDIR "lib"
+    CACHE STRING "Desura Lib Dir")
+set(DATADIR ""
+    CACHE STRING "Desura Data Install Dir")
+set(DESKTOPDIR "/usr/share/application"
+    CACHE STRING "Desktop installation direcoty")
```
Jookia: direcoty

Jookia:
I haven't tested it, but it looks good. You said this provokes #190. Maybe this should be investigated?

build_out/run.sh
karolherbst, line 13: this file is renamed by cmake.

Jookia, line 13: I mean from run.sh to something else. It'd be a bit confusing for new coders.

src/static/util/code/UtilOs.cpp
karolherbst, line 144: I moved it out to UtilOs_cmake.cpp.inc because we have to configure this path via cmake (only cmake knows, where the data files are installed and how)

build_out/run.sh
karolherbst, line 13: maybe something like: "precomfigured_desura_launch_script_cmake.sh"?

karolherbst:
I will try them out tomorrow while I am on the way to work

karolherbst:
I want to merge this on monday, so it should be tested deeply. I don't want to have bugs because of this merge

Mailaender:
works: https://build.opensuse.org/package/show?package=Desurium&project=home%3AMailaender

karolherbst:
Okay, seems to work, I will merge this on Monday

PR #308, opened 2012-09-09
usercore: escape \n in desktop files

### LINKED ISSUE #306: Desura generates invalid desktop shortcuts if the game's description contains newlines
1. Install [AlterEgo](http://www.desura.com/games/alterego) (or, I'd assume, another game with a multi-paragraph description)
2. Use "Create Desktop Shortcut" in the Play list context menu.
3. Note that the resulting .desktop file appears as an ordinary file on the desktop.
4. Examine the content and notice that the value for "Comment=" spans multiple lines without quoting.

I'm not sure which specific revision I'm running, but the PPA's .deb version is 2012.08.23-makson1~ppa1

karolherbst:
we should escape a new line in the comment field of desktop files. Further we should check, if the genre is empty.

fixes #306

### DIFF
--- a/src/shared/usercore/code/ItemHandle_nix.cpp
+++ b/src/shared/usercore/code/ItemHandle_nix.cpp
@@ -316,8 +316,10 @@ void ItemHandle::installLaunchScripts()
 
 inline gcString createDesktopFile(ItemInfoI* i)
 {
+	using boost::algorithm::replace_all_copy;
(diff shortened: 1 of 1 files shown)


TheSiege:
compiles and works with alter ego

karolherbst:
the problem is, that the description part of alter ego was updated in the meantime, so this error doesn't occur without this changes

karolherbst:
I want to merge this in 24 hours

PR #310, opened 2012-09-11
cmake: use gcc options as default

karolherbst:
sometimes the compiler identification can not be determinied. In that case, we will use the gcc configuration


karolherbst:
I want to merge this in 24 hours

PR #311, opened 2012-09-11
add possibility to build cef separately from desurium

karolherbst:
this will safe a lot of compile time, because cef won't change in general. Further it will make updates on gentoo much faster.

Please be aware, that the CEF build still depends on the desurium build system and the desurium build system still needs a cef archive for header files.


TheSiege:
works; ebuild works; please merge it saves a lot of build time

karolherbst:
I want to merge this on 2012-09-15T14:00+00:00

PR #312, opened 2012-09-12
implement Cmake version handling

### LINKED ISSUE #266: how to handle versioning in desurium
We should talk about how to handle versioning inside desurium.

For me a branching system (like in mesa or Linux) won't make sense for me, because we don't want to "support" old versions and just rely on the newest one.

As anagromataf said in IRC, the "wine"-system would be a good one. So we would have a "stable" version-branch and an "unstable" version-branch at the same time. So we would try out things in the unstable one and push them into the stable one after a while.

Do you have any other ideas or aspects?

### LINKED ISSUE #277: Get rid of the version file
Currently some version data is read from a file found along side the executable, which should be unnecessary because the file doesn't hold enough data to justify it, and reading from files takes an unnecessary amount of time.
Currently I don't know exactly what these version numbers are used for, except as a replacement to Microsoft Windows registry keys, used by MySQL database code.
The version data is gathered from the functions GetAppBuild() and GetAppBranch(), which are found in src/static/util/code/UtilLinux.cpp

karolherbst:
fixes #277
fixes #266

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,30 @@
+version 0.8.0:
+--------------
+
+fixed issues:
+ * Add create menu shortcut on linux (https://github.com/lodle/Desurium/issues/29)
+ * Tooltips (https://github.com/lodle/Desurium/issues/31)
+ * List Games & CDKeys (https://github.com/lodle/Desurium/issues/32)
+ * Upload & Download issue (https://github.com/lodle/Desurium/issues/33)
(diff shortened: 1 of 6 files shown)


karolherbst:
I want to merge this before branching. You should take a look over the AboutForm.

TheSiege:
everything looks fine and as it should; just the picture is mising from the page

TheSiege:
ok regression fixed; ready for merge IMO

karolherbst:
I want to merge this in a few minutes and branch for release after it

PR #313, opened 2012-09-12
replace run.sh with desura

cerebrux:
there is no run.sh in install/ instead there is desura that launches the application

### DIFF
--- a/README
+++ b/README
@@ -44,7 +44,7 @@ curl (while building without ares support)
 sqlite
 tinyxml
 
-Finally run Desura using ./install/run.sh
+Finally run Desura using ./install/desura
 
 If you're on a 64bit system and want 32bit support, abort building it and go to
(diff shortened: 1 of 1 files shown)


Cheeseness:
The problem is that run.sh sets some environment stuff. Distro maintainers can ensure that Desurium is built with whatever flags it needs to not need that stuff set, but so far as our work goes (and in the interests of keeping things flexible), I think we need to continue to use run.sh.

karolherbst:
run.sh is renamed by cmake within the installation step and installed in BINDIR.
the compiled desura binary will be installed into RUNTIME_LIBDIR.

This is important, because we need to set some env variables before launching desura. A direct desura call will fail. But the run.sh should be stay as run.sh so we won't get confused within our repository, becauce in fact it is a run.sh

karolherbst:
ahh you are right, I forgot to edit the README because of this change

thx for figure this out

PR #316, opened 2012-09-16
Bug259

### LINKED ISSUE #259: cef: reduce features like cups support
so we need less dependencies and it will decrease build time

karolherbst:
fixes #259


TheSiege:
it compiled and worked

PR #317, opened 2012-09-16
Remeoved mkdir "package"

cerebrux:
There is no need to create "package" folder. If user creates the "package" then cd's in the package folder and finaly runs the command to create the package, he/she will get an error from Cmake. 
Instead either just run the command on the root folder of the source code or add the code in the "package" folder e.g.:
1. mkdir package
2. cd package
3. extract here the source code
4. run cmake -DPACKAGE_TYPE="{PACKAGE_TYPE}" -DINSTALL_DESKTOP_FILE=ON -DCMAKE_INSTALL_PREFIX="/opt/desura"
5. make package

### DIFF
--- a/README
+++ b/README
@@ -74,10 +74,8 @@ Create Packages
 
 You can simply run the following commands to create packages for installing desura with your package manager:
 
- 1. mkdir package
- 2. cd package
- 3. cmake -DPACKAGE_TYPE="{PACKAGE_TYPE}" -DINSTALL_DESKTOP_FILE=ON -DCMAKE_INSTALL_PREFIX="/opt/desura" ..
- 4. make package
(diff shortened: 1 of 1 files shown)


karolherbst:
no, we are using out of source buids, and this should be the normal behaviour with cmake.

maybe you forgot the ".." at the end?

In addition to that, your third point is invalid.
1. download sources
2. extract sources
3. cd into sources
4. create out of source build dir
5. cd into out of source build dir
6. run cmake .. $options...
7. make $goal

cerebrux:
You are right... damn I didn't se the ".." ! Please close this request as invalid, and sory for any inconvinience :-)

karolherbst:
invalid

PR #327, opened 2012-10-04
Bug324

### LINKED ISSUE #324: 32 bit games do not start on 64 bit system.
This seems to be regression. In August, when I build Desurium with -D32BIT_SUPPORT I was able to play 32 bit games. Now (at least since 09.09.2012) I cannot do that. I get error:

11:31   Failed to launch item.: You need 32 bit libraries to play this game [75.0]

Tested on Trine 2 and Sacred Gold. If I start games manually from the install directory, they work fine.

karolherbst:
fixes #324


makson96:
I have tested it.
This fix the issue.
I have not seen any regression (so far).

PR #332, opened 2012-10-15
Bug121

### LINKED ISSUE #121: "Exit" just hides the window
On both the official Desura release for Linux and the GPL build, both the logo-button "Exit" command and the tray icon "Exit" command usually (but not always) only minimize Desura to the tray.

This bug happens so often (probably 9 times out of 10) that I have taken to quitting Desura with `killall desura` without even bothering to try the Exit option.

I also suspect an interaction between this and Desura's SQLite commit behaviour might be why the list of recently-played games is never correct. (My best guess is that Desura only commits changes either on a long timer or on exit, and I tend to play until it's coursework time, then kill Desura immediately to avoid temptation.)

karolherbst:
should fix #121 read b42bed043f7a383944b9a8b274c6f38d7bfeb823 for more information

### DIFF
--- a/src/shared/usercore/code/ItemManager.cpp
+++ b/src/shared/usercore/code/ItemManager.cpp
@@ -1057,16 +1057,23 @@ void ItemManager::parseLoginXml2(TiXmlElement* gamesNode, TiXmlElement* platform
 
 	ParseInfo pi(UM::ItemInfoI::STATUS_ONACCOUNT, 0, false, &maps);
 
-	XML::for_each_child("platform", platformNodes, [this, &pi](TiXmlElement* platform)
+	try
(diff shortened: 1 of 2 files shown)


Jookia:
Some testing seems to confirm that this fixes #121. However, without the patch I've only managed to make #121 happen once.

TheSiege:
ok; so I can't reproduce #121 all the time; though sometimes it happens and this commit slashes the time needed a lot. Compiles fine and everything; doesn't seem to break stuff
tested on gentoo; ubuntu 12.04 fedora 17

karolherbst:
I think I will add another issue related to this one, but it won't be a "bug"

PR #333, opened 2012-10-15
Version number 0.8.0-devel

makson96:
The 0.8 milestone is not yet achieved. So this change version number from 0.9.0-devel to 0.8.0-devel.


karolherbst:
master is 0.9.0-devel
0.8 branch is 0.8.0-devel

PR #343, opened 2012-10-28
separate compiler and linker flags

### LINKED ISSUE #336: separate compiler and linker flags
mentioned here: https://bugs.gentoo.org/show_bug.cgi?id=395837#c22

karolherbst:
fixes #336


TheSiege:
compiles, but I am not sure how to test it otherwise

karolherbst:
compile test is enough

PR #346, opened 2012-10-30
Bug329

### LINKED ISSUE #329: User avatar not respecting proper alpha
The client seems to only support 1-bit alpha in avatars in the top left of the window (next to the menu button). It ignores semi-transparent pixels and applies full alpha to them.

karolherbst:
fixes #329

candidates for stable branch:
11447a1bb08e6ff0873708b1177ebf4ffb18714b


karolherbst:
as lodle mentioned I've restored the else clause for windows

TheSiege:
compiles and looks like it works on linux

PR #347, opened 2012-11-03
Spanish translation updated

lugonu:
I have updated the Spanish translation for Desurium. Previously,  the xml format was broken and it included some strings in a different language.

Karol: You may also receive these files by email, they are the same version.


karolherbst:
okay, seems to work. Thanks

PR #351, opened 2012-11-13
Ebuild: add missing dependencies

karolherbst:
As I created my Chroot env on a vserver these deps were missing.


TheSiege:
so ... not much changes; emerge --depclean gives nothing new to remove; and it is not a build change ... so you can merge

PR #352, opened 2012-11-14
Wx detect

makson96:
If someone want to use system wxWidgets with Desurium he would probably have default 2.8 version and 2.9.3 as a choice. But if 2.9.3 is not default, than even if wxWidgets_CONFIG_EXECUTABLE and wxWidgets_wxrc_EXECUTABLE is set, the test will fail because those options are not passed to the test. This pull is fixing the issue.


karolherbst:
seems okay, I will looker deeper into it later

TheSiege:
compiles fine; doesn't break things (tried ebuild; build_desura.sh and a few cmake options)

karolherbst:
We could change the ebuild a bit with this changes

TheSiege:
I am pretty sure wx.eclass handles that in gentoo ...

karolherbst:
makes sense

karolherbst:
works for me

PR #355, opened 2012-11-19
threads: we need another joinable check

### LINKED ISSUE #348: desurium not working with boost-1.52
desurium is currently not working with boost-1.52 as the main windows does not show up and the process crashes;

```
*** Caught unhandled unknown exception; terminating
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::thread_resource_error> >'
  what():  boost thread: thread not joinable: Invalid argument
```

karolherbst:
this should fix #348 fully


TheSiege:
fixes the issue with installing things; just takes a while;

PR #358, opened 2012-11-23
util: don't sanitize '..' to '.'

### LINKED ISSUE #239: Styling Issue - Scrollbar colours
This should be a trivial bug fix, make the scrollbars the same colour on all sections, looks odd on the play section to have a black scrollbar than all other tabs grey?

Just a styling issue.

karolherbst:
and some tests are added

also fixes #239

### DIFF
--- a/src/static/util/code/UtilString.cpp
+++ b/src/static/util/code/UtilString.cpp
@@ -225,7 +225,11 @@ std::string sanitizeFilePath(const std::string &str, char slash)
 		}
 
 		if (str[x] == '.' && x+1<size && str[x+1] == '.')
+		{
+			out += "..";
+			x++;
 			continue;
+		}
 
 		out += str[x];
 	}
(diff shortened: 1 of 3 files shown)


TheSiege:
fixes issue; works

lodle:
The whole reason that .. are removed to stop bad url's accessing paths that they shouldnt be allowed to

karolherbst:
This makes sense, but isn't sandboxing a better solution for that?

lodle:
well, given unlimited time, yes but desura does load content from 3rd parties.

karolherbst:
I meant only sandboxing the internal browser. But in general, if we want to be FHS complianet, we have to respect that, that we have to access files, which aren't subfolders if the "binary folder" of desura.

I could also eliminate all ".." in the theme path folder, where this path is coming from.

lodle:
Go for it then, im offering insight as to why it was like that, if needs change, then the code has to change.

karolherbst:
I will create an issue because of sandboxing.

PR #367, opened 2012-11-29
wx_controls: remove private API usage

karolherbst:
we should not use private fields or private APIs in general.


TheSiege:
still compiles on my gentoo

karolherbst:
I want to do some VC checks before merging this, but I won't have enough time for it until Tuesday

Jookia:
This seems like a fantastic addition.

karolherbst:
seems to compile fine on msvc100 (well, I was able to compile the whole project on msvc100, except cef and every MFC dependent part)

PR #368, opened 2012-11-29
common: use template paramters in AutoDeleteLocal

karolherbst:
we have to add this to be more c++ standard compliant.

also fixes compile errors on mingw

### DIFF
--- a/src/common/Common.h
+++ b/src/common/Common.h
@@ -386,22 +386,22 @@ template <typename T>
 class AutoDeleteLocal : public AutoDelete<T>
 {
 public:
-	AutoDeleteLocal(T* t) : AutoDelete(t)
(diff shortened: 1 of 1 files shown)


TheSiege:
still compiles on my gentoo

karolherbst:
I want to merge this on Sunday

karolherbst:
no build failure with visual studio, too

Jookia:
Does this exhibit the same behaviour as the previous iteration of it? Prefixing that stuff with the parent class name looks a little weird. Is there a reason why that's done?

karolherbst:
this code was only used with msvc, but on windows mingw, which is gcc in fact, complained about this. The error is not the missing class prefix, but the missing template argument.

karolherbst:
compiles fine on msvc

PR #371, opened 2012-12-10
cmake: rename BR to BRANDING_DIR

karolherbst:
BR is an ugly name an nobody knows what the use is. We should always use understandable names.


TheSiege:
still compiles on my gentoo

karolherbst:
I want to merge this on Sunday

Jookia:
Looks fine to me.

PR #372, opened 2012-12-10
cmake: don't link against Boost_LIBRARIES

karolherbst:
We should link against every single lib we want use and not against all.

This will allow us to use a builtin boost library for tests.


TheSiege:
works on my gentoo

karolherbst:
I want to have build and run test from different distributions. Ubuntu, ArchLinux, Suse, Fedora would be nice.

r-ick:
Builds and runs OK here in Arch.

PR #377, opened 2012-12-16
Update build_out/data/language/hungarian_utility.xml

AllenWalkerHU:
Fully retranslated "english_utility.xml" to Hungarian.

### DIFF
--- a/build_out/data/language/hungarian_utility.xml
+++ b/build_out/data/language/hungarian_utility.xml
@@ -1,30 +1,66 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
-<lang name="Magyar (Hungarian)" skip="true">
+<?xml version="1.0" encoding="UTF-8"?>
+<lang name="Magyar (HU)" skip="true">
 	<author>
-		<username>NSD4rKn3SS</username>
-		<profile>http://www.desura.com/members/nsd4rkn3ss</profile>
+		<username>AllenWalkerHU</username>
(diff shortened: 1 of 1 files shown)


Jookia:
Does this account for https://github.com/lodle/Desurium/wiki/Translation-TODO ?

AllenWalkerHU:
Really sorry but I don't get what this question means.

Jookia:
Ah, I mean that since the previous translations were done, new things have been added that aren't in the translations. They're in that list there. Does your new translation include the changes?

AllenWalkerHU:
Oh, now I see. Well, they aren't included YET, since I've just registered here about half an hour ago after I finished translating that file and didn't really had time to check everything here. But I'll update every one of what I missed.

karolherbst:
Is this patch tested against our current master/0.8 branch? Because it would make sense to merge this also into the 0.8 branch.

karolherbst:
closed because of #379

AllenWalkerHU:
What's wrong with this one? I don't think I missed or mistranslated any string. Also checked it with the 0.8 branch and it's ready to be merged.

Jookia:
I'm a little confused too.

karolherbst:
see #379

karolherbst:
ohh my fault, I thought these both were the same file

AllenWalkerHU:
I thought we have a bigger problem here. No worries anyway. And thanks!

PR #378, opened 2012-12-16
uicore: fix compile error on msvc100

karolherbst:
msvc100 doesn't know constexpr yet. But we can safely replace constexpr with static const


TheSiege:
compiles fine here

PR #384, opened 2012-12-30
update dependency to 4.6 wrt Range-based for | N2930 | GCC 4.6

TheSiege:
There was a lot of trouble when people tried building desurium with gcc-4.5 on gentoo; I think we narrowed that done to 'Range-based for | N2930 | GCC 4.6'


karolherbst:
Wie should drop  this dependency and do a check against active GCC version seen here: https://github.com/shiznix/unity-gentoo/blob/master/unity-base/unity/unity-5.18.0.ebuild#L62

TheSiege:
not in the ebuild; all needed dependencies have to be listed; as gcc is slotted there is no problem
https://github.com/shiznix/unity-gentoo/blob/master/unity-base/unity/unity-5.18.0.ebuild#L47
but checking also against the active gcc is a good idea

karolherbst:
And you should tag the commit with "ebuild:"

karolherbst:
But here is more work to do, also in portage. What if you want to build desurium with another compiler...

TheSiege:
seperate compilers would be controlledd over a useflag ( see clang useflag in eix); but I think we should aim to have one working ebuild in the gentoo tree before we experiment with making the ebuild good for different compilers

karolherbst:
it is working for now, that's why I am worried about a pkg_pretend check.

I would like to have something like this:

if gcc && gcc version >= 4.6
|| if clang && clang version >= 3.0
|| if icc && icc version >= ??

hasufell:
https://gist.github.com/d81b1f108601880a5a56

karolherbst:
but how does "gcc-*-version" interact with other compilers? But I like the idea more than having an explicit dependency on gcc-4.6.

Or should we backport to gcc-4.5 instead? I don't know how common 4.6 is today. We should create an official dependency list we want to stick with.

karolherbst:
I will open an issue for that

hasufell:
https://gist.github.com/d81b1f108601880a5a56/2d3d3060d7c33dcf54855bb5e10e447fcc313512

karolherbst:
this looks good. So we could also do some fancy stuff like: "if [[ $(tc-getCC) =~ clang ]]; then" and other version checks?

It would be awesome having multiple compiler support directly in portage, but I think that we should have something like that until 2014

karolherbst:
Would be nice if you would fix #388 with this pull request. I would replace all the usage with the result.

PR #399, opened 2013-01-13
PKGBUILDs: desurium: make check: run ./build/src/tests/UtilLinux_test. (...

matthiaskrgr:
...commented out since failing currently)


karolherbst:
we won't disable test only because they are failing, fixing is the right way to go

matthiaskrgr:
I would have commented this in as soon as the tests were fixed. Meh

karolherbst:
Please fix the test instead ;)

PR #402, opened 2013-01-13
travis: major rework.

matthiaskrgr:
upate the system before compiling anything
comment out tests temporarily since they are failing currently
make clean after build is done
dump status report into #desura irc channel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: cpp
 compiler: gcc
 before_install:
-  - sudo apt-get install libnotify-dev libtinyxml-dev libv8-dev libboost-dev libboost-date-time-dev libboost-filesystem-dev libboost-thread-dev libboost-system-dev libboost-test-dev
+  - sudo apt-get update
+  - sudo apt-get upgrade
+  - sudo ./install-deps.sh
   - wget "http://ppa.launchpad.net/fransschreuder1/usbpicprog-stable/ubuntu/pool/main/w/wxwidgets2.9/libwxgtk2.9-dev_2.9.3-1_i386.deb"
   - wget "http://ppa.launchpad.net/fransschreuder1/usbpicprog-stable/ubuntu/pool/main/w/wxwidgets2.9/wx2.9-headers_2.9.3-1_i386.deb"
   - wget "http://ppa.launchpad.net/fransschreuder1/usbpicprog-stable/ubuntu/pool/main/w/wxwidgets2.9/libwxgtk2.9-0_2.9.3-1_i386.deb"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,7 +1,9 @@
 language: cpp
 compiler: gcc
 before_install:
-  - sudo apt-get install libnotify-dev libtinyxml-dev libv8-dev libboost-dev libboost-date-time-dev libboost-filesystem-dev libboost-thread-dev libboost-system-dev libboost-test-dev
+  - sudo apt-get update
+  - sudo apt-get upgrade
+  - sudo ./install-deps.sh
```
karolherbst, line 6: update and upgrade are only needed for newer deps, which we don't need. We should avoid high build times as much as possible on travis-ci

.travis.yml
```diff
@@ -18,11 +20,23 @@ before_script:
 script:
   - cmake .. -DBUILD_CEF=off -DWITH_ARES=off -DFORCE_SYS_DEPS=on -DDEBUG=on
   - make -j4
-  - src/tests/UtilLinux_test 
-  - ctest --output-on-failure .
+#  - src/tests/UtilLinux_test
+#  - make test
+#  - ctest --output-on-failure .
+  - make clean
```
karolherbst, line 26: we don't need "make clean" on travis because the vm is thrown away after build

.travis.yml
```diff
@@ -18,11 +20,23 @@ before_script:
 script:
   - cmake .. -DBUILD_CEF=off -DWITH_ARES=off -DFORCE_SYS_DEPS=on -DDEBUG=on
   - make -j4
-  - src/tests/UtilLinux_test 
-  - ctest --output-on-failure .
+#  - src/tests/UtilLinux_test
+#  - make test
+#  - ctest --output-on-failure .
```
karolherbst, line 25: please enable the tests, because we should always check for compile AND for runnings tests, otherwise it would be very senseless to use travis-ci

matthiaskrgr:
further questions:

```
why "make clean" ?
why "sudo apt-get update sudo apt-get upgrade sudo ./install-deps.sh"
```

we should avoid stuff which will only increase build time

also the calling of the install-deps script is heavier than installing stuff, which is missing.

I thought that under some weird circumstances "make clean" could fail as well.
Additionally I tried to compile cef after makeing clean desurium, but that didn't work for some reason.

We first update the system in order to,well, have an updated build enviroment.  The travis env is only updated once per month iirc.

regading install-deps, hey, this was the initial reason why I made this patch and you commented with "nice idea :) " :P
I want to make sure that install-deps works properly, if it misses some deps, build shall fail!

.travis.yml
```diff
@@ -18,11 +20,23 @@ before_script:
...
 notifications:
+  irc:
+    channels:
+      - "irc.freenode.org#desura"
+    template:
+      - "%{repository} (%{commit}) by %{author} : %{message} %{foo} "
+      - "Build details: %{build_url}"
+      - "Latest commit: %{compare_url}"
+    on_success: change
+    on_failure: always
```
karolherbst, line 36: wouldn't this pollute the IRC channel? I mean if I would work on one of my branches and push things every day...
we are getting informed of pull requests, where you can read the travis-ci results and because we only push through "pull requests", we don't need this really

.travis.yml
```diff
@@ -18,11 +20,23 @@ before_script:
...
   recipients:
-    - desurium@karolherbst.de
+  #  - desurium@karolherbst.de
+    - matthias.krueger@famsik.de
```
karolherbst, line 39: I think we should delete this part

karolherbst:
okay calling install-deps.sh should be okay, because we would only update stuff, we really need.

calling apt-get update/upgrade is to much for me, because an update once a month is more than enough for a reliable build environment

.travis.yml
matthiaskrgr, line 36: Well we can set on_success: change     on_failure: change   which will mean we _ONLY_ get "spam" if the build switches from good to broken  or from broken to good again.

.travis.yml
matthiaskrgr, line 25: alright

matthiaskrgr:
I think that when we enable travis for lodle/Desurium, pull requests are automatically tested (pre merge). I thought this could be interesting.

PR #405, opened 2013-01-13
Travis:

matthiaskrgr:
Use install-deps.sh
Run make test
Notify in #desura, if build status changes from broken to ok, or from ok to broken. (otherwise not)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ before_script:
 script:
   - cmake .. -DBUILD_CEF=off -DWITH_ARES=off -DFORCE_SYS_DEPS=on -DDEBUG=on
   - make -j4
-  - src/tests/UtilLinux_test
-  - make test
   - ctest --output-on-failure .
 notifications:
   irc:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -18,11 +18,16 @@ before_script:
 script:
   - cmake .. -DBUILD_CEF=off -DWITH_ARES=off -DFORCE_SYS_DEPS=on -DDEBUG=on
   - make -j4
-  - src/tests/UtilLinux_test 
+  - src/tests/UtilLinux_test
+  - make test
```
karolherbst: ctest --output-on-failure is "better" than "make test", because with ctest we are able to put additional arguments to the ctest runner. (make test will invoke ctest only)

.travis.yml
```diff
@@ -18,11 +18,16 @@ before_script:
 script:
   - cmake .. -DBUILD_CEF=off -DWITH_ARES=off -DFORCE_SYS_DEPS=on -DDEBUG=on
   - make -j4
-  - src/tests/UtilLinux_test 
+  - src/tests/UtilLinux_test
```
karolherbst: this line should be deleted, because ctest will print the error. This was my mistake.

PR #407, opened 2013-01-17
optimize graphics using matthiaskrgr/image_optim

matthiaskrgr:
stats: 148 binary files changed 278933 b (272 kb) -> 159998 b (156 kb) => [-118935 b (-116 kb)]


karolherbst:
I don't really know what this jpeg and png commands do in your script.
1. we should keep licensing information within the picture files.
2. How does it effect runtime performance? Safing RAM is good, thought it won't be that much, because the bitmaps will be decompressed anyway (you can't display png pictures directly, you need the RGBA values anyway).

karolherbst:
Okay I found no picture errors, so for me it is okay, if the question above are clarified.

karolherbst:
With the amount of data I don't think anybody will notice a drop in runtime performance, but I will give some guys the change to check this.

karolherbst:
We should check the copyright stuff with the desura icons before merging this.

karolherbst:
okay I looked into the pngs and found no copyright information. Also most of the pictures were converted into greyscale pngs. The result of it is the drop of the PLTE table (pallete of the png file).

Issue #408, opened 2013-01-18
fix workaround introduced in 3a317cf


karolherbst:
3a317cf5201efb96469434b4f320971ae256c544 I used a workaround, I should fix this.

karolherbst:
workaround had fixed this commit: 1f343f1ea858a652716599474d64dc691c0ab60d

Issue #362, opened 2012-11-27
Upload Bug - Exit on completion


Protektor-Desura:
There is now a bug in the current code that I downloaded that as soon as an MCF is finished uploading the client exits. Doesn't matter the size of the file. This did not happen in the version you guys wanted me to test when you thought you had the upload download bug fixed and it was working correctly then. Something in the code has changed since then that is effecting uploads.

karolherbst:
does not happen here. I am able to upload mcf after creation and client does not exit

karolherbst:
but it happens with the binaries from the here: http://nemoder.homenet.org:8080/desurium-0.8-RC3-i686-Nemoder.tar.gz

```
*** glibc detected *** /home/karol/Applications/desurium/desurium-0.8-RC3-Nemoder/lib/desura: double free or corruption (!prev): 0x0a1219f8 ***
======= Backtrace: =========
/lib32/libc.so.6(+0x771a2)[0xf67491a2]
/home/karol/Applications/desurium/desurium-0.8-RC3-Nemoder/lib/libstdc++.so.6(_ZdlPv+0x1f)[0xf69234bf]
/home/karol/Applications/desurium/desurium-0.8-RC3-Nemoder/lib/libuicore.so(+0x215ee0)[0xf560bee0]
/home/karol/Applications/desurium/desurium-0.8-RC3-Nemoder/lib/libwx_gtk2u_desura-2.9.so.3(_ZN12wxWindowBase7DestroyEv+0x39)[0xf6f8c659]
/home/karol/Applications/desurium/desurium-0.8-RC3-Nemoder/lib/libwx_gtk2u_desura-2.9.so.3(_ZN12wxWindowBase15DestroyChildrenEv+0x2b)[0xf6f8c69b]
/home/karol/Applications/desurium/desurium-0.8-RC3-Nemoder/lib/libwx_gtk2u_desura-2.9.so.3(_ZN8wxWindowD1Ev+0x79)[0xf6e52fc9]
/home/karol/Applications/desurium/desurium-0.8-RC3-Nemoder/lib/libuicore.so(+0x72ce1)[0xf5468ce1]
/home/karol/Applications/desurium/desurium-0.8-RC3-Nemoder/lib/libuicore.so(+0x1c41b1)[0xf55ba1b1]
[...]

PR #410, opened 2013-01-18
wx_controls: fix workaround

karolherbst:
this reuqest fixes a workaround in 1f343f1ea858a652716599474d64dc691c0ab60d introduced #362.
Instead of creating a new instance of an incompatible class I fixed the height in gcULProgressBar directly.

This bug was caused, because gcULProgressBar had a uint8_t field, which resulted into some errors within glibc.

This pull request will fix both #362 and #408


karolherbst:
I want to create another RC after this one got merged

TheSiege:
works

PR #414, opened 2013-01-23
partially revert "Travis:"

karolherbst:
the update of all needed package can fail unter some circumstances.
See here: https://travis-ci.org/karolherbst/Desurium/builds/4325989

We should install only the missing ones


karolherbst:
this seems okay for me, because it will fix the error. And if not, then the error will occur less

PR #419, opened 2013-01-24
cmake: mac os X requires cmake 2.8.10

### LINKED ISSUE #418: Mac: CMake versions prior to 2.8.10 may fail to download external projects
The official binaries of CMake prior to 2.8.10 may fail to download files over https. This is due to the version of libcurl being linked against CMake having SSL disabled. Either the minimum required version should be increased or this bug should be documented somewhere (eg in the build instructions).

karolherbst:
prior versions are shipped without SSL support (used by ExternalProject)

fixes #418


karolherbst:
doesn't seem to break anything.

PR #421, opened 2013-01-26
add findlib.sh

### LINKED ISSUE #370: findlib.sh missing
In Linux the Desura client installs libraries markes as dependencies for game even if those already exists in system. I have verified this in Debian Squeeze  and Xubuntu 12.10 and with both Desura and Desurium built from git master today. Only tested with my own game (The Platform Shooter), which has  libGLU and libXaw set as dependencies.

After building Desurium in Xubuntu 12.10 and installing only this one game from there the libraries can be found in these locations:

/home/mikko/.desura/games/14729/3496/lib/libXaw.so.7
/home/mikko/.desura/tools/202/libXaw.so.7
/usr/lib/i386-linux-gnu/libXaw.so.7
/usr/lib/x86_64-linux-gnu/libXaw.so.7

/home/mikko/.desura/games/14729/3496/lib/libGLU.so.1
/home/mikko/.desura/tools/95/libGLU.so.1
/usr/lib/i386-linux-gnu/libGLU.so.1
/usr/lib/i386-linux-gnu/libGLU.so.1.3.1
/usr/lib/x86_64-linux-gnu/libGLU.so.1
/usr/lib/x86_64-linux-gnu/libGLU.so.1.3.1

This is a major problem for my game, because the libXaw.so.7 library installed by Desura is incompatible with all Debian versions (and many distributions based on them). See forum post here: http://www.desura.com/groups/desura/forum/thread/problems-with-libxawso7-installed-by-desura

EDIT (karolherbst):
we need the findlib.sh shell script for this

karolherbst:
fixes #370 

NOTE: this is a candidate for the stable branches

### DIFF
--- a/build_out/findlib.sh
+++ b/build_out/findlib.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+BIT=""
+
+if [ x"$2" == x"64" ]; then
+	BIT="64"
+	LIBDIRS="/lib64 /lib/x86_64-linux-gnu /lib /usr/lib64 /usr/lib/x86_64-linux-gnu /usr/lib /usr/local/lib /usr/lib/x86_64-linux-gnu/nss /usr/lib/nss"
+else
+	BIT="32"
+	LIBDIRS="/lib /lib/i386-linux-gnu /lib/i686-linux-gnu /lib32 /usr/lib /usr/lib32 /usr/lib/i686-linux-gnu /usr/lib/i386-linux-gnu /usr/local/lib /usr/lib/i386-linux-gnu/nss /usr/lib/i686-linux-gnu/nss /usr/lib/nss"
+fi
+
+
+for DIR in $LIBDIRS; do
+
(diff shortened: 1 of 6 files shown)


build_out/findlib.sh
```diff
@@ -0,0 +1,38 @@
...
+
+for DIR in $LIBDIRS; do
+
+	if [ -e "$DIR/$1" ]; then
+
+		BUILD=`file -L "$DIR/$1" | grep 64-bit`
+
+		#check to see if 64 bit lib
+		if [ x"$BIT" == x"64" -a -z "$BUILD" ]; then
+			continue;
```
matthiaskrgr, line 22: "continue;"  ?

karolherbst, line 22: "for DIR in $LIBDIRS; do". I did not write this file, so I don't know. Everything I know is, that this script seems to work

karolherbst:
got response "yes findlib.sh is allowed to be used"

PR #422, opened 2013-01-26
implement "check" (make check) into build_desura.sh

matthiaskrgr:


build_desura.sh
```diff
@@ -21,5 +27,12 @@ if [ ! -d "build" ] ; then
...
 cd build
 cmake .. -DCMAKE_INSTALL_PREFIX=$PREFIX -DBINDIR=$BINDIR -DRUNTIME_LIBDIR=$LIBDIR -DDATADIR=$DATADIR -DBUILD_CEF=OFF || exit 3
-make install $@ || exit 4
+make install $args || exit 4
+
+if [[ "$check" == "true" ]] ; then
+	printf "Running 'make check'\n"
+	# already in /build
+	make test
+	./src/tests/UtilLinux_test
```
karolherbst: Is there any reason why you are calling this test explicitly?

karolherbst:
This patch is breaking if you passing more options to this script:

```
$ ./build_desura.sh -j8 check
Make sure to run sudo ./install-deps.sh before compiling!

We are compiling CEF first
-- using clang
-- detected 64bit
-- Configuring done
-- Generating done
-- Build files have been written to: /home/karol/Dokumente/repos/Desurium/build_cef
make: *** No rule to make target `check'.  Stop.
make: *** Waiting for unfinished jobs....
```

build_desura.sh
matthiaskrgr: If it is included in   make test  I can skip it.
I had in mind that not all tests were included in make test.

matthiaskrgr:
Mmh, I tried to avoid this :/

karolherbst:
tests are declared here: https://github.com/lodle/Desurium/blob/master/src/tests/CMakeLists.txt
so if one of the test wouldn't be included, it wouldn't be build anyway

karolherbst:
but now it works fine

Issue #423, opened 2013-01-27
Add an option to bundle wxgtk (at least until 3.0 got released)


TheSiege:
I had a little chat at #gentoo-games on freenode; it might be very reasonable to introduce an option to specifically bundle wxgtk-2.9 (whichever is the latest working version); as 2.9 is a development version it can frequently breake compatibility and people might get really upset when suddenly the new wxgtk broke desurium compilation (or functionality) and therefore could not update to a new version of wxgtk.
another proposed option was to maybe statically link it
- [x] remove wxGTK dependency from CEF
- [x] update ebuilds and PKGBUILD

karolherbst:
wxGTK can't be statically linked, because it will lead into really big troubles. Bundling would be the way to go.

karolherbst:
And because we don't patch wxGTK on Linux anymore, it won't conflict with the system wide installed one (if there is any)

karolherbst:
maybe it is possible on Linux, because I tried static linking only on windows.

karolherbst:
http://trac.wxwidgets.org/wiki/Roadmap

"Next stable release: 3.0"

TheSiege:
if so; then please provide it as a default option, but leave the option to unbunble it

TheSiege:
anagromataf: please test this: https://github.com/karolherbst/Desurium/commits/bug423
Works;

PR #425, opened 2013-01-29
cef: drop wxWidgets dependency

karolherbst:
needed for #423

updated ebuilds will come later this day

NOTE: this series is a candidate for the stable branch


TheSiege:
works

PR #427, opened 2013-01-29
Correct shell path

mark7:
[[ ]] syntax is not understood by dash, which is what /bin/sh runs on
Debian.  Use bash instead.

### DIFF
--- a/build_desura.sh
+++ b/build_desura.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


karolherbst:
I don't think, that we should just move to /bin/bash. We should better replace [[ with [ and change the stuff needed for it.

matthiaskrgr:
Alternative fix:  https://github.com/lodle/Desurium/pull/430

PR #432, opened 2013-01-30
clean up cef.patch a bit

karolherbst:
no functional change within this series, but it will reduce size of the cef patches.

This should make a port to newer versions of cef easier.


TheSiege:
works

PR #436, opened 2013-01-30
fix typo

matthiaskrgr:


karolherbst:
mhh, google:

"Get source from" About 268,000 results
"Get sources at" About 48,000 results

Jookia:
IMHO it should be 'Get the sources from:'

karolherbst:
google says everything is all right! :P

PR #437, opened 2013-01-30
rename mfc_upload_cli to mcf_upload_cli since the binary says "MCF".

matthiaskrgr:


matthiaskrgr:
this should proly get merged into 0.8 as well

PR #441, opened 2013-01-31
Pkgbuild 

matthiaskrgr:


karolherbst:
please try to remove this commit: 46cd23d0ee476b19ebaecb597a28d751f9d47690
I don't see any reason, why we should add this

PR #443, opened 2013-01-31
Adhere to OpenDesktop menu-spec

Mailaender:
If the `.desktop` file is invalid according to http://standards.freedesktop.org/desktop-entry-spec/1.0/ it will break rpm on http://build.opensuse.org

```
ERROR: No sufficient Category definition: /home/abuild/rpmbuild/BUILDROOT/Desurium-0.8.0_rc4-0.i386//usr/share/applications/desura.desktop 
WARNING: file extension in Icon line: /home/abuild/rpmbuild/BUILDROOT/Desurium-0.8.0_rc4-0.i386//usr/share/applications/desura.desktop
ERROR: Icon file not installed: /home/abuild/rpmbuild/BUILDROOT/Desurium-0.8.0_rc4-0.i386//usr/share/applications/desura.desktop (../../share/desurium/desura.png)
WARNING: Empty GenericName: /home/abuild/rpmbuild/BUILDROOT/Desurium-0.8.0_rc4-0.i386//usr/share/applications/desura.desktop
Errors in installed desktop file detected. Please refer to http://en.opensuse.org/SUSE_Package_Conventions/RPM_Macros
error: Bad exit status from /var/tmp/rpm-tmp.4K6kkj (%install)
```

This should fix it.

### DIFF
--- a/cmake/scripts/desura.desktop
+++ b/cmake/scripts/desura.desktop
@@ -1,8 +1,11 @@
 [Desktop Entry]
-Name=Desurium
+Version=1.0
 Type=Application
-Comment=Free software version of Desura game client
-Exec=@BIN_INSOURCE_DIR@/desura
-TryExec=@BIN_INSOURCE_DIR@/desura
-Icon=@DATA_INSOURCE_DIR@/desura.png
-Categories=Game;
+Name=Desurium
+GenericName=Digital games distribution client
+Comment=Free software version of Desura digital distribution client
+Exec=desura
+TryExec=desura
+Icon=desura
+Categories=Game;Amusement;
+Terminal=false


karolherbst:
What is wrong about: "Categories=Game;"?

karolherbst:
According to this: http://standards.freedesktop.org/menu-spec/menu-spec-1.0.html#category-registry

Categories=Game;Amusement;

is not good, because a desktop environment doesn't need to accept this.

Further desktop-file-validate is fine with this.

karolherbst:
In Addition to that, you should use absolute paths while building for system wide installs. Using relative paths with cmake here is just insane, because it doesn't fit to the needs (Do you want to search relative from a path or using a fixed system path?)

Mailaender:
I think http://en.opensuse.org/openSUSE:Packaging_desktop_menu_categories is the way it is because it works better with KDE. The path in .desktop files are not relative, but default to the search paths like `/usr/bin` defined by the system.

karolherbst:
yes, but the Category=Game; is valid xdg standard. I just want to use here a value, that should be understandable on every system. If OpenSUSE is not accepting this, it is either a bug within OpenSUSE or devs are forced to add an entry.

But because a desktop environment doesn't need to accept "Category=Amusement;" it will fail on other distributions.
1. in the log I read "../../share/desurium/desura.png" so I thought that "Icon=../../share/desurium/desura.png" was set.

Maybe you should upload the generated file (if possible). If you would set something like "-DDATADIR=/usr/share/games/" then the value for Icon should be "/usr/share/games//desura/desura.png". Though it isn't that nice to use absolute paths, it is still valid against the xdg specification.

If you want to perfectly fit your distribution, you should create your own desktop file. This desura.desktop file is just a try to create a file that is valid against the xdg specification

Mailaender:
Okay I will then use RPM macros or `desktop-file-install` to fix this up.

PR #449, opened 2013-02-04
Clean CMakeLists.txt from testing wx_widgets version

makson96:
It seems that there is no more test for WX Widgets 2.9.3. So the reference to the test should be removed from CMakeLists.txt.


karolherbst:
I don't think we should remove this check yet, because it will invoked if you use system wxGTK (what you could still do)

makson96:
You are right. This test should be only deleted when wxWidgets 2.9.4 will be stable with Desurium.

makson96:
But maybe if someone deliberately pass -DFORCE_BUNDLED_WXGTK=OFF, than we should allow him to test any wxWidget 2.9 he want (not only 2.9.3).

karolherbst:
But then the test should be removed as well: [cmake/tests](../tree/master/cmake/tests)

PR #456, opened 2013-02-09
build: use sourceforge as wxwidgets download source.

matthiaskrgr:
The current mirror was failing for me since the last 12 hours.


matthiaskrgr:
note:  this is a candidate for 0.8 branch

karolherbst:
I think this is for normal user the best choice, because package maintainer will use pre downloaded archives anyway

PR #458, opened 2013-02-09
npwrapper bugfixes

karolherbst:
this should eliminate linker errors on some systems (for example ARCH)

NOTE: this series is a candidate for the stable branch


matthiaskrgr:
Fixes build problems on arch linux.

karolherbst:
I think this should fix most of the errors now, seems to work on gentoo, arch and ubuntu.

PR #461, opened 2013-02-10
npwrapper: link wrapper against glib

### LINKED ISSUE #459: desura crash since #485
/usr/games/lib64/desura/desura: symbol lookup error: /usr/games/lib64/desura/libdesura_flashwrapper.so: undefined symbol: g_strdup_printf

then crash

adding glib-2.0 to set(PLATFORM_LIBS dl) doesn't fix it

karolherbst:
fixes #459


TheSiege:
works and fixes crash

PR #462, opened 2013-02-10
Install script

matthiaskrgr:


build_desura.sh
```diff
@@ -1,43 +1,142 @@
...
+	echo "Making clean for cef..."
+	if [ -d "build_cef" ] ; then
+		cd ./build_cef/
+		make clean
+		cd ../
+	fi
+	if [ ! -d "build" ] ; then
+		echo "Removing install directory..."
+		rm -rf "install"
+	fi
```
karolherbst: - else: find install/ -type f -iname libcef_desura.so -delete ?

build_desura.sh
```diff
@@ -1,43 +1,142 @@
...
+	echo "Making clean for desurium..."
+	if [ -d "build" ] ; then
+		cd ./build/
+		make clean
+		cd ../
+	fi
+	if [ ! -d "build_cef" ] ; then
+		echo "Removing install directory..."
+		rm -rf "install"
+	fi
```
karolherbst: same as above:
else
  find install/ -type f  ! -iname libcef_desura.so -delete
?

karolherbst:
I think it would be better to make commands and subcommands.

```
./build_desura.sh compile desurium
./build_desura.sh compile cef rebuild desurium
```

matthiaskrgr:
I'll have a look at this but it seems a bit more complicated and so I'd like to have this merged first :P

matthiaskrgr:
> I think it would be better to make commands and subcommands.
> 
> ./build_desura.sh compile desurium
> ./build_desura.sh compile cef rebuild desurium

can't be done.

make fails to accept passed -j values such as "-j 2" (-j2 works) or  "-j2 check"

make: the `-j' option requires a positive integral argument

matthiaskrgr:
unfinished progress in case someone finds a workaround: http://pastebin.com/qHH08MJn

PR #467, opened 2013-02-11
Pkgbuild

matthiaskrgr:


distro/archlinux/git/PKGBUILD
```diff
@@ -54,7 +48,7 @@ build() {
   cd build
   msg "Running cmake..."
   CMAKE_ARGS="
-    -D32BIT_SUPPORT=$MULTILIB
+    -D32BIT_SUPPORT=1
```
karolherbst, line 51: you should add here a check against target arch. But I think on 32bit this flag is without effect anyway

distro/archlinux/release/cef/PKGBUILD
```diff
@@ -6,15 +6,17 @@ pkgver=0.8.0_rc7
...
+license=('BSD')
 arch=('i686' 'x86_64')
 makedepends=('cmake' 'git' 'boost' 'subversion' 'scons' 'gendesk' 'sed')
 depends=('v8' 'tinyxml' 'sqlite' 'boost-libs' 'bzip2' 'gtk2' 'openssl'
          'curl' 'c-ares' 'libnotify' 'nss' 'libgnome-keyring' 'bison'
          'cups' 'flex' 'alsa-lib' 'libjpeg-turbo' 'libxpm' 'libx11'
          'gconf' 'xdg-user-dirs' 'flac' 'libpng' 'speex' 'zlib' 'gperf'
-         'xdg-utils' 'libwebp' 'libevent' 'libxslt' 'yasm' 'libxxf86vm')
+         'xdg-utils' 'libwebp' 'libevent' 'libxslt' 'yasm' 'libxxf86vm'
+         'binutils-multilib' )
```
karolherbst: cef doesn't need multilib stuff

distro/archlinux/release/desurium/PKGBUILD
```diff
@@ -55,7 +49,7 @@ build() {
   cd build
   msg "Running cmake..."
   CMAKE_ARGS="
-    -D32BIT_SUPPORT=$MULTILIB
+    -D32BIT_SUPPORT=1
```
karolherbst: same as above

karolherbst:
I think the best would be to provide an additional package named desurium-32bit-support.

This desura_bittest binary can be a shell script returning 0 as well. You have only to make sure, that there is 32bit support on the system, which can be easily done by the package manager.

The current problem I see is, that sometimes you don't want to run 32bit software and don't have a multilib system.

matthiaskrgr:
OK, I'll 32bit pkgbuilds.

PR #470, opened 2013-02-12
install script DO NOT MERGE

matthiaskrgr:
DO NOT MERGE
this is a test, I'm trying to implement the features suggested here: https://github.com/lodle/Desurium/pull/462#issuecomment-13395573

but I want to make sure that this works for other people as well.

input:
./build_desura.sh --compile desura -j 2 --rebuild cef check --help nope

output should be 
Make sure to run sudo ./install-deps.sh before compiling!

compile desura -j 2 
rebuild cef check 
help nope

when I only use --help at the end, I get "help help help" on last line, I have to fix this.

### DIFF
--- a/build_desura.sh
+++ b/build_desura.sh
@@ -8,8 +8,11 @@ echo $@ | sed -e 's/\-\-/\n/g' > $filename
 IFS=$'\n'
 set -f
 for i in $(<$filename); do
-	echo $i | awk '{print $1}' # action
-	echo $i | awk '{print $2,$3}' #whatever
+	action=`echo $i | cut -d\  -f1`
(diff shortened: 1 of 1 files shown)


matthiaskrgr:
@karolherbst  :)

karolherbst:
I think the best would be to parse the arguments first and then do the actions

and what should happen with this:

```
./build_desura.sh --compile desura -j 2 --rebuild desura check --help nope
```

matthiaskrgr:
I just wanted to know if the code above works at all!

karolherbst:
I think this is just too complicated

matthiaskrgr:
I'll try anyway :>

PR #463, opened 2013-02-10
Gentoo bug fixes

karolherbst:
this should all bugs on gentoo systems (found by @TheSiege)

PR #472, opened 2013-02-12
ebuild: filter --as-needed out of linker flags

karolherbst:
as a workaround we will have this until it is fixed


karolherbst:
was tested within #463

PR #474, opened 2013-02-14
Set CURL to build with /MD instead of /MT

ghost:
All other libraries are built with /MD and mixing the two causes issues under VS.


karolherbst:
good catch

Issue #243, opened 2012-06-12
cmake: check for all needed dependencies


karolherbst:
some dependencies are not checked by cmake:
- patch
- automake
- compiler checks:
  - gcc-4.5
  - clang-3.1

feel free to report more deps (I will update this list)

Mailaender:
https://build.opensuse.org/package/binary?arch=i586&filename=Desurium-1346150676-96.1.i586.rpm&package=Desurium&project=home%3AMailaender&repository=openSUSE_12.1 shows the requirements after build and https://build.opensuse.org/package/view_file?file=Desurium.spec&package=Desurium&project=home%3AMailaender lists all the build dependencies. However I don't know which belong to dependencies (only for wxWidgets) as I just added them blindly whenever make crashed.

Jookia:
Is this still a thing?

karolherbst:
yeah

PR #475, opened 2013-02-15
add FindPatch.cmake

ghost:
I've added FindPatch.cmake. This is only used when compiling under MSVC. CMake will exit with an error if patch.exe is not found (it'll search in system paths and the default installation directories of msysGit). I've also removed tortoiseGit as it's unneeded (msysGit provides both Git and patch.exe).

This pull needs to be tested on Linux/mingw as I made some changes that could affect these platforms.


karolherbst:
I like most part of this request, but I am worried about using patch only from a git installation. It should be possible to build desurium on Windows without any git installation (because sometimes you will only download a versioned archive)

ghost:
CMake will check %PATH% before checking in the git installation. The user is not affected by this change if they already have patch.exe in their path.

cmake/scripts/Patch.bat
```diff
@@ -1,4 +1,4 @@
-patch -p0 -N -i %1
+"${PATCH_EXECUTABLE}" -p0 -N -i %1
```
karolherbst: I would suggest to use @PATCH_EXECUTABLE@ here and configure with @ONLY 

I think it can be confusing to have some sh semantics within a batch file. @...@ indicates clearly, that it is cmake related

karolherbst:
I think we should also check for patch on Linux, because of #243

ghost:
It seems extremely unlikely that a developer would have all the necessary dependencies installed to compile Desurium except for patch.

karolherbst:
it is still a direct dependency and we should check all of them.

FindPatch.cmake should work for UNIX based system as well, shouldn't it? We don't need to configure patch.sh as well, but we should fail if it isn't in PATH

ghost:
Ok, change has been made.

karolherbst:
seems fine to me now, thanks

CMakeLists.txt
```diff
@@ -175,6 +175,14 @@ if(WIN32)
   find_package(PythonLibs REQUIRED)
 endif()
 
+find_package(Patch REQUIRED)
+if(MSVC)
+  configure_file(${CMAKE_SCRIPT_PATH}/patch.bat ${CMAKE_BINARY_DIR}/scripts/patch.bat @ONLY)
```
karolherbst: I think all generated files should be stored in ${CMAKE_GEN_SRC_DIR}.
Right, some generated files aren't put there, but we should move all to this directory

karolherbst:
last comment and then everything should be fine

ghost:
Didn't even realise CMAKE_GEN_SRC_DIR existed. It should be all good now.

karolherbst:
thanks

PR #478, opened 2013-02-16
Default 32bit support on 64bit Linux

makson96:
I guess that most gamers on 64 bit system want their 32 bit games to work by default. install-deps.sh already provides all dependencies for building this support (at least on Ubuntu and Debian).

### DIFF
--- a/README
+++ b/README
@@ -46,10 +46,6 @@ tinyxml
 
 Finally run Desura using ./install/desura
 
-If you're on a 64bit system and want 32bit support, abort building it and go to
-the 'build' folder, and edit the CMake cache as to enable 32BIT_SUPPORT.
-(You can use ccmake or cmake-gui for this).
-
 Build Desura on Windows
 =====================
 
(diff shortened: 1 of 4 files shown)


matthiaskrgr:
Hmm, perhaps we should rather implement an 32bit option into build_desura.sh ?

makson96:
I can do that, but I am not in favor of it. If we add to many options into build_desura.sh it will be no different, than just directly invoking  cmake.

Nevertheless I think that 32bit support should be enabled by default, because all gamers do expect it.

karolherbst:
I think this will lead into the same problems, but the other way around (try 32bit build on non multilib systems).

I think the best solution would be to enable this only by default if and only if the compiler supports multilib

makson96:
This test work for me. It will make 32 bit support enable only if test for compiling code with C flag "-m32" (on 64 bit system) is passed.

karolherbst:
Isn't there an easier way to find this out?

makson96:
I haven't found any by Google :P

If you have any better idea please let me know and I can use it.

karolherbst:
did anybody test this on a non multilib 64bit system?

makson96:
Well, I did.

On Ubuntu after uninstalling "gcc-multilib" test fails and 32 bit support is turned OFF. After installing package back and running cmake once again test is successful and 32 bit support is ON.

Mailaender:
Requested for openSUSE on https://bugzilla.novell.com/show_bug.cgi?id=805424.

PR #481, opened 2013-02-17
Add pack_deb/pack_rpm option to build_desura.sh

makson96:


build_desura.sh
```diff
@@ -62,6 +62,20 @@ clean_desurium() {
...
+	if [ $PACKAGE = "DEB" ]; then
+		echo "Building DEB package..."
+	elif [ $PACKAGE = "RPM" ]; then
+		echo "Building RPM package..."
+	fi
+	if [ ! -d "build" ] ; then
+		mkdir build
+	fi
+	cd build
+	cmake .. -DPACKAGE_TYPE=$PACKAGE -DINSTALL_DESKTOP_FILE=ON -DCMAKE_INSTALL_PREFIX="/opt/desura" || exit
```
matthiaskrgr, line 71: This won't compile cef which is compiled by default.

@karolherbst  how should this be handled?

makson96, line 71: Before Pull Request I tested it on my local machine and it does build CEF, wxWindgets and all.

BUILD_CEF is set to ON by default and if it is not changed by passing argument to cmake, than cef will build together with desurium.

matthiaskrgr, line 71: Ok, sorry, then I was wrong :)

karolherbst:
you should use a new build dir for that (maybe build_package?). Otherwise you will get conflicts in the cmake cache, because you already configured the project with other options (which could lead into undefined behaviour like non installed CEF)

makson96:
What do you think about it now?

karolherbst:
looks good to me now

PR #489, opened 2013-02-21
VS library fixes

ghost:
This fixes most issues with building third party libraries under VS. There are two issues still remaining

Some of the archive files contain symlinks. CMake under windows can not deal with symlinks and will exit with an error. We have two options here:
1. Repackage files with tar -h. This will ensure that symlinks are replaced with a copy of the file they point to. This can not be done for breakpad as it has an absolute symlink pointing to autotools. We can probably just delete this symlink however
2. Provide binaries of these libraries in our repo. This is probably the best solution for VS users as they don't need to download/compile these libraries and it'll simplify some of the cmake scripts. The fact that we need to patch these libraries complicates things somewhat.

npwrapper will not compile under VS. I haven't looked into npwrapper yet so I'm not sure how difficult this is to fix (or if we even use npwrapper under Windows).


karolherbst:
All of it seems fine to me. I left a comment on 2a2ce23349d901429ffb50977d9a9111a0267fd1  though.

We should also think about a better solution in b831ca6380de26141bf1dd2c1a76b1c10ec791a2. I am sure this patch will fail under different versions of vc, because of the values in  "PreprocessorDefinitions"

ghost:
I could read the file in. Use STRING(REGEX REPLACE ...) to find and replace any occurrences of MultiThreaded with MultiThreadedDLL and then write the file back out. That might be more robust.

karolherbst:
I would guess that, too. But I am feeling a bit worried about that, because I don't know how stable breakpad will run with shared runtime, but a REGEX replace should be the better solution

ghost:
Ok, all issues should be resolved now. Thanks for the hint regarding rebasing in git.

I'm not exactly sure if I put breakpad-VS-patch.cmake in the right folder but I couldn't think of anywhere better to put it.

ghost:
I forgot to mention that switching to debug in the IDE will break the project. This is fixable but I'd prefer to concentrate on getting Desurium to build under VS Release before tackling that issue.

karolherbst:
I know this issue, but yes building is more important now

PR #490, opened 2013-02-22
manager: change global objects to singletons

### LINKED ISSUE #190: No language loaded.

karolherbst:
fixes #190 

every fallthrough code path should be eliminated here.

GetString was returning the key as long as the LanguageManager was NULL. But this cannot happen anymore. This request should also fix an error with a not loaded theme (strange colors)

EDIT: this is a candidate for the stable branch

### DIFF
--- a/src/include/managers/Managers.h
+++ b/src/include/managers/Managers.h
@@ -109,11 +109,11 @@ public:
 	virtual void getConCommandList(std::vector<ConCommand*> &vList)=0;
 };
 
-ThemeManagerI*		GetThemeManager();
-LanguageManagerI*	GetLanguageManager();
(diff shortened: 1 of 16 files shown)


Jookia:
This gets my vote regardless if it fixes #190; It simplifies a lot, and as far as I can see it looks good.

karolherbst:
@Jookia I said something in this direction to Cheesness, too. I think it is much easier to use references and thread safe creation of global objects than hoping that a pointer is fully initialized and that one thread will be faster than every other.

Mailaender:
Tried it 10 times and could not reproduce #190. Looks like you fixed it.

PR #494, opened 2013-02-23
Update README

makson96:
This pull request will update and simplify installation instruction under Linux in README file.


README
```diff
@@ -17,38 +16,48 @@ For more information, visit http://www.moddb.com/groups/desura/forum/thread/open
...
+- cmake (2.8.5 or later)
+- PkgConfig (for building DEB and RPM packages)
+- Boost (date_time, filesystem, thread system, test_exec_monitor, unit_test_framework)
+- Freetype
+- GTK2
+- libX11
+- libEvent
+- libNotify (0.7 or later)
+- GLib2
+- OpenSSL
```
karolherbst: OpenSSL is only required if we are building curl

karolherbst:
It would be nice to have these squashed together into one commit

PR #505, opened 2013-03-01
add h.264 support, but disable it by default

karolherbst:
There is some patent and license stuff, so be aware of that before redistributing cef with h.264 support

v2:
- add option to remove flash support
- updated ebuilds
- documentation for H264_SUPPORT

NOTES:
- this is a candidate for the stable branch
- a full rebuild of cef is needed
- to test this you can simply remove all flash binaries


matthiaskrgr:
Will this change anything regarding dependencies?

karolherbst:
not yet, maybe I find some libs I can take from the system later

matthiaskrgr:
https://www.archlinux.org/packages/extra/i686/x264/ something like this maybe?

TheSiege:
works great, i'd say merge

TheSiege:
before merging, plase get rid of the flash dependency in desurium ebuild, switch it to desurium-cef and make it a dep if bindist is enabled

karolherbst:
yeah, nice sugestion

TheSiege:
so; everything seems in order; how about merging this and the language fix and creating a new rc?

Issue #389, opened 2013-01-03
desurium-cef: build system does not respect CFLAGS


karolherbst:
found here: https://bugs.gentoo.org/show_bug.cgi?id=395837#c22

PR #513, opened 2013-03-16
respect toolchain

hasufell:
also set V=1 which forces verbose build output, respect MAKEOPTS and
MAKE/CMAKE_BUILD_TOOL instead of hardcoded "make"

this should mostly fix bug https://github.com/lodle/Desurium/issues/389


hasufell:
see https://github.com/lodle/Desurium/pull/517

PR #518, opened 2013-03-16
improve generated desktop file

karolherbst:
with this patches it should be possible to install the cmake generated desktop file.


Mailaender:
:+1:

karolherbst:
I want to use domenu in the ebuild instead of installing it via cmake.

I will change this today or tomorrow

Mailaender:
Does Gentoo not support http://freedesktop.org/wiki/Software/desktop-file-utils?

karolherbst:
on gentoo, there are plattform independent tools (which will use desktop-file-utils on Linux)

PR #520, opened 2013-03-17
findlib.sh: improvements

karolherbst:
posix compliance, use ldconfig to get additional LIBDIRS,
also respect LIBDIRS env variable


karolherbst:
seems to work for me, but the first run takes some time, but not much

PR #527, opened 2013-03-26
add basic Mac OS X support

karolherbst:
I worked a bit on compile compatibility with Mac OS X.

Here is the result.

I also change some handling with the binary type of binaries and moved some util methods from UTIL::LIN to UTIL::OS and UTIL::UNIX (new UTIL namespace for UNIX based systems).

Some methods aren't implemented for Mac OS X, but they will print out a warning if called.

Currently nearly nothing works, though you can launch desurium and the login screen will come up.

Also the 10.7 SDK is required.


Jookia:
I don't think this should be merged in to master.

karolherbst:
@Jookia right, but you can test this patches and it will stay on my branch. I could push this branch to lodle if you want.

karolherbst:
I could also split this into more pull request to separate trivial things from complex ones

japboy:
First and great step to support OSX :+1:

Jookia:
Why closed?

karolherbst:
I removed my branch, because I have a newer one https://github.com/karolherbst/Desurium/tree/apple2

PR #529, opened 2013-03-27
fixes Bug515

### LINKED ISSUE #515: desura:// links don't work from external browser if client is already opened
I can reproduce it with Firefox 19.0.0.2 and Desurium RC 9:
1. Fire up Desurium
2. Go to http://www.desura.com/games/openra
3. Click http://www.desura.com/games/openra/play
4. Click desura://launch/games/openra
5. Choose "Start application: Desura"

`Desurium is already running, If it's not, remove "~/.desura/cache/applock".`

It should execute the action without requiring the client to be not already running.

karolherbst:
fixes #515 

@Jookia do you know why did you do https://github.com/karolherbst/Desurium/commit/5dd2abf4c7667e2619c1dbe1029c1183791497b7 ? It displayed an error message everytime you want to launch a game from the desura website


Jookia:
I assumed that was the end of the application and some IPC was done with the main Desura. My bad.

karolherbst:
it could be, that there was a bug in pre wxGTK-2.9.3, because we are using wxSingleInstanceChecker. The applock file is removed after a crash of derusium for me.

I am not sure, if there are any other possibilities to launch desura/desurium while another instance is crahsed, but marked as running (maybe a zombie process?)

PR #530, opened 2013-03-27
use systemwide xdg-open and clean up URL whitelist

### LINKED ISSUE #531: desurium is launching urls in the external browser
current master of desurium will open links in the system web browser if gnome-open is not installed on the system.

with this pull request https://github.com/lodle/Desurium/pull/530 it will also happen if xdg-open is not installed or the call fails (very unlikely)

karolherbst:
desura used to bundle all xdg tools in the binary distribution. But we want to use the system wide installed ones

fixes #531

NOTE: fe8ff527c36ece7384a0329470e3e0932bcc8c63 and 828541ab52122eef548a3d76048d7c4e400bf076 are candidates for the stable branch

### DIFF
--- a/src/executable/bootloader_lin/code/DesuraMain.cpp
+++ b/src/executable/bootloader_lin/code/DesuraMain.cpp
@@ -108,7 +108,7 @@ void ShowHelpDialog(const char* msg, const char* url, const char* type)
 	
 	if (url)
 	{
-		std::string systemString("xdg/xdg-open ");
+		std::string systemString("xdg-open ");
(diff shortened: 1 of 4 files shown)


Mailaender:
:+1:

Mailaender:
Don't forget to check if xdg-utils are installed.

karolherbst:
I am sure, that xdg-utils is one of the packages which is installed on any desktop distributions. Also if xdg-open and gnome-open is failing, desurium will print an error.

PR #535, opened 2013-04-05
cmake: add mcf_util

karolherbst:
I don't know how I could forgot this, but here is it


karolherbst:
I want to merge this before a new rc

Jookia:
No objections here.

karolherbst:
I will check if we have to link against mcfcore

PR #539, opened 2013-04-13
Fixed Boost copy_file error.

### LINKED ISSUE #533: mcf_upload_cli fails to compile on arch with boost 1.53.0
[ 99%] Built target crashdlg
[ 99%] Built target desura_bittest
Linking CXX executable mcf_upload_cli
../../shared/usercore/libusercore.so: undefined reference to `boost::filesystem::detail::copy_file(boost::filesystem::path const&, boost::filesystem::path const&, boost::filesystem::copy_option, boost::system::error_code_)'
collect2: error: ld returned 1 exit status
make[2]: *_\* [src/tools/mcf_upload_cli/mcf_upload_cli] Error 1
make[1]: **\* [src/tools/mcf_upload_cli/CMakeFiles/mcf_upload_cli.dir/all] Error 2
make: **\* [all] Error 2

Jookia:
This fixed #533. For real.

### DIFF
--- a/src/static/util_fs/code/UtilFs.cpp
+++ b/src/static/util_fs/code/UtilFs.cpp
@@ -348,7 +348,7 @@ void copyFile(Path src, Path dest)
 	try
 	{
 		bf::remove(bf::path(dest.getFullPath()));
-		bf::copy_file(bf::path(src.getFullPath()), bf::path(dest.getFullPath()));
+		bf::copy(bf::path(src.getFullPath()), bf::path(dest.getFullPath()));
 	}
 	catch (bf::filesystem_error e)
 	{


karolherbst:
mhh, copy is doing implicit more than copy_file. If we want to make sure to copy a file, we should use copy_file. If we don't care about what type of "file" it is (symlink, file, directory), we can use copy

Jookia:
Copy runs copy_file if it's a file. We could add a path type check if needed. I don't know of another way around it, but my fix shows that there's some weird linker stuff going on.

karolherbst:
yes, I don't think we should "change behavior" here, but fix the linker issue directly. We could add Boost_FILESYSTEM_LIBRARY to the usercore and mcfcore library list, but I won't do that, because we don't use this interface there directly, but via another (static) dependency like util and util_fs.
That's why I want to know where this linking error is comming from. But maybe we should move all of this static libs to shared libs aynway (to reduce time needed to find the cause of such errors and to reduce binary size)

karolherbst:
I think we should not allow it to copy directories in a function name copyFile.

karolherbst:
@Jookia how do you like this fix #542? It will pretend the same error with other methods using scoped enums in boost (copy_file is currently the only one, but you knows which guy wants to use one of these methods and wondering about a stupid linking error)

Jookia:
That's a great fix.

karolherbst:
So I am closing this request

PR #541, opened 2013-04-17
js/playList: parse rating to float

### LINKED ISSUE #374: 10.0 rated games not sorted correctly
I bought overlight and if I sort my game list by rating, it is bellow 7.1 and not at the top :(

Screenshot from 2012-12-11 23:30:30

karolherbst:
the custom sort function is comparing strings. String comparision in JavaScript is fine as long as both strings have the same lenght and the decimal mark at the same position.

But "10.0" was the only case violating this. "10.0" < a is returning false for all a's matching ^[0-9].[0-9]$

fixes #374

v2: we need to handle the rating coloumn explicitly, also remove this dirty eval hack with an associative array access

### DIFF
--- a/build_out/data/themes/default/js/play.list.js
+++ b/build_out/data/themes/default/js/play.list.js
@@ -69,10 +69,14 @@ sortList = function(list) {
 
 		for(item in list) {
 			try {
-				temp[item] = {key: item, value: eval("list[item]."+sortFieldCache+"()")};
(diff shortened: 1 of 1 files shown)


karolherbst:
okay, this broke every other sorting

karolherbst:
regression fixed

TheSiege:
works as expected

PR #547, opened 2013-05-04
add option to disable breakpad and disable it on non official builds

### LINKED ISSUE #297: remove breakpad
we don't really need google-breakpad for desurium so we should get rid of this dependency in the OSS client and only compile related stuff when it is for the official Desura client.

karolherbst:
in development

fixes #297


karolherbst:
I will close this, because I want to add more stuff to it

PR #549, opened 2013-05-05
npwrapper: add debian flash plugin path

karolherbst:


karolherbst:
there are still some issues with disabled h.264 support, I will investigate later

PR #551, opened 2013-05-21
uicore: fix Breadcrumbs height

### LINKED ISSUE #213: Breadcrumbs are positioned too high.
I've managed to get this fixed somehow by removing the last column, specifically changing [this line](https://github.com/lodle/Desurium/blob/master/src/shared/uicore/code/HtmlToolBarControl.cpp#L72) to having 10 columns instead of 11. But that removes the last search element.

karolherbst:
fixes #213


karolherbst:
I have the same issue like Jookia, so I have to find the cause

Jookia:
This is something that's been bugging- Oh, you posted as I was writing. Anyways, nice patch. I couldn't figure it out.

karolherbst:
so I reverted your change, so the "/" are right, but not the home button

Jookia:
So the home button is bugged?

karolherbst:
yeah I have to investigate a bit further on this

karolherbst:
I will close this, because it isn't finished yet

PR #555, opened 2013-05-29
Made README a markdown file.

WhyNotHugo:
Content is the same, readability with text-based applications is identical, but github renders this way more nicely. :)


karolherbst:
looks good to me, thanks

PR #574, opened 2013-07-11
cmake: download files from sourceforge

### LINKED ISSUE #544: where should archvies for building desurium be stored?
we can't upload new files on github so we have to search for another download server.

The uploaded archives are broken for windows, so we have to find a solution for that fast.

karolherbst:
fixes #544


TheSiege:
works as expected

PR #575, opened 2013-07-14
Bug491

Smilex:
Replaces tinyxml with tinyxml2.


karolherbst:
some comments without a look into the changed files:
- you should remove all tinyxml2 files which are not tinyxml2.cpp or tinyxml2.h. We don't need the others.
- you should remove all unreleated commits via git rebase -i desura/master
- you should handle the tinyxml2 stuff like tinyxml was handled: new target, include only if system ones not found, set variables in tinyxml2 cmake file also if not found systemwide

Smilex:
Did I get it right this time?

third_party/tinyxml2/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_definitions(-D_LIB)
+add_library(tinyxml2 STATIC tinyxml2.cpp)
+
+set(TINYXML_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include/ PARENT_SCOPE)
```
karolherbst: should be set(TINYXML_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR} PARENT_SCOPE)
because the header file isn't in a include subdirectory

karolherbst:
this little diff should be applied on this branch. It is only some documentation, install_deps.sh and travis stuff: http://bpaste.net/show/114420/

karolherbst:
as I found out, travis is ubuntu 12.04 so it doesn't have a tinyxml2 dev package. So we have to remove it from the apt-get install command and remove "-DFORCE_SYS_DEPS=on" from the cmake call.

In addition to that, there are some tiny little things in the diff, I will make comments there

CMakeLists.txt
```diff
@@ -75,6 +75,7 @@
 # TODO: silent output for externel dependencies
 project(Desura)
 
+
```
karolherbst, line 78: two empty lines, we should have at max one (non critical)

third_party/tinyxml2/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_definitions(-D_LIB)
+add_library(tinyxml2 STATIC tinyxml2.cpp)
+
+set(TINYXML_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR} PARENT_SCOPE)
```
karolherbst, line 4: the pkg_config call is setting TINYXML_INCLUDEDIR, so we should use this here too

src/include/Event.h
```diff
@@ -46,6 +46,7 @@
 
 #include <vector>
 #include <algorithm>
+#include <cassert>
```
karolherbst, line 49: Is there a reason why we have to include cassert here?

src/include/XMLMacros.h
```diff
@@ -298,13 +295,10 @@ inline uint32 processStatus(TiXmlDocument& doc, const char* root)
 	return v;
 }
 
-#include <tinyxml.h>
-
-inline void loadBuffer(TiXmlDocument& doc, char* buff, size_t buffLen,
-	TiXmlEncoding encoding = TIXML_ENCODING_UTF8)
+inline void loadBuffer(tinyxml2::XMLDocument& doc, char* buff)
```
karolherbst, line 298: we should try to have a range checked buffer here. I don't know the interface of doc.Parse, but if the buffer is a template, we could use the boost::autobuffer stuff or something similiar

src/shared/usercore/code/InstalledWizardThread.cpp
```diff
@@ -398,7 +398,7 @@ void InstalledWizardThread::parseItems1(TiXmlNode *fNode, WildcardManager *pWild
 		if (!gameId.isOk())
 			return;
 
-		TiXmlElement* info = NULL;
+			tinyxml2::XMLElement* info = NULL;
```
karolherbst, line 401: little whitespace issue

src/shared/usercore/code/ItemManager.cpp
```diff
@@ -1195,7 +1195,7 @@ void ItemManager::parseModsXml(UserCore::Item::ItemInfo* parent, ParseInfo &pi)
 		if (!internId.isOk())
 			return;
 
-		TiXmlElement* infoNode = NULL;
+			tinyxml2::XMLElement* infoNode = NULL;
```
karolherbst, line 1198: whitespace issue

karolherbst:
I went through all the changes an I might  found everything, but I will take a new look after the corrections

src/include/Event.h
Smilex, line 49: It suddenly gave me a complier error about assert not being defined, so I just included it, which fixed it

src/include/XMLMacros.h
Smilex, line 298: int tinyxml2::XMLDocument::Parse( const char \*  xml)
Only version of the Parse function.

karolherbst, line 298: so buf is a NULL terminated xml char array. Mhh this could cause security problems, because parsing a xml is triggered from the outside (internet).

Jookia, line 298: If we're wrapping it why not use std::string or gcString?

karolherbst:
I think I will clean up this request and merge it via cli (without removing your author tags of course)

src/include/XMLMacros.h
Smilex, line 298: I don't see how wrapping it as a std::string or gcString would matter. It's a matter of getting the string from a size read or not. But I also don't see how a premature null termination in the XML can cause security issues.

karolherbst:
it would be nice if you would update your branch with my cleaned up commits found here: https://github.com/karolherbst/Desurium/commits/bug491

karolherbst:
closed due newer pull request

PR #578, opened 2013-08-07
wx_controls: fix #329 also for win32 and remove the alpha argument

### LINKED ISSUE #329: User avatar not respecting proper alpha

karolherbst:
I left the code for windows, because I wasn't able to test it there. But now I checked it and it fixes the alpha problem there, too.


karolherbst:
@lodle is this "!!alpha" string used for anything?

karolherbst:
I will close this and put this commit together with the win32 pull request

Issue #577, opened 2013-08-04
Upcoming Games - Release date


makson96:
As you can see on the picture, something is wrong with displaying release dates of the games:
https://imageshack.us/a/img163/7018/eztl.png
It says:
"Dominions 4: Thrones of Ascension - Available in {0} 28 {0} days {1} hour."
So those zeros and brackets seems to be some kind of malfunction.

lodle:
The {0} are place holders for content. Looks like your language file is out of date or incorrect. Try with English?

makson96:
Yep, in English it looks fine. I will tweak translation files than and look for the solution and propose patch.

karolherbst:
thanks a lot :)

PR #580, opened 2013-08-08
Bug577

makson96:
This pull request fix issue #577 for all supported languages. Note that I was mainly using Google Translator for languages other than Polish.

### DIFF
--- a/build_out/data/language/esperanto.xml
+++ b/build_out/data/language/esperanto.xml
@@ -564,14 +564,13 @@
         <str name="#IS_PAUSED">Paŭza</str>
         <str name="#IS_PREORDER_STATUS">Antaŭmendo</str>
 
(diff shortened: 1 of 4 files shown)


karolherbst:
I don't think we should make translation with Google Translator. It would be better to let people do it, who are able to speak the languages.

makson96:
Well, I know Russian good enough to see it is correct. For Finish and Esperanto those are only minor changes which seems ok to me (the singular form looks very similar to plural, which was previously present). But if you want me to revert Finish and Esperanto anyway, let me know.

karolherbst:
I don't have a strong mind about Esperanto anyway. I will leave some notes in the commit

karolherbst:
no, it seems okay

PR #587, opened 2013-08-20
patch/chromium-nss-3.15: check against major version as well

karolherbst:
we should do it this way, just to be more clean


Jookia:
This applies right?

karolherbst:
somehow I screwed my build directory... I will give the result later, though I already checked it...

karolherbst:
so it works for me

PR #600, opened 2013-09-06
clean up bootloader and don't fork before execution also wxgtk debug build fix

### LINKED ISSUE #585: Debugging is hard.
Right now when running Desura under GDB, it's impossible to break in a file, break in GDB at runtime, catch crashes, or catch exceptions.

karolherbst:
fixes #585

this forking lead to some missbehave in gdb, also it doesn't make much sense.

I also cleaned some unused methods and simple single line function away, so that this class should be more straight forward.


Jookia:
@lodle what do you think?

karolherbst:
@Jookia does this request help the bug?

Jookia:
It fixes the bug. Merging.

Issue #509, opened 2013-03-07
dual license desurium to BSD+GPLv3


karolherbst:
Desura Pty Ltd. wants desurium to be dual license under BSD and GPLv3. The BSD clause should be only valid for desura ltd. and GPLv3 for anybody else.

@matthiaskrgr @makson96 @jcnix @harmonise @boskee @Smilex @sadam36 @Cheeseness @yurikoles @TheSiege @LarryRotgut @cerebrux @AllenWalkerHU @naryl @Telemaniaka @madrang and @ssokolow you have contributed to desurium and so we need your thoughts on that.

agrees:
- Cheeseness
- Jookia
- makson96
- yurikoles

wants other license:
- Cheeseness
- karolherbst
- makson96
- ssokolow
- TheSiege

does not care:
- harmonise
- jcnix
- matthiaskrgr
- Telemaniaka

no answers yet:
- boskee
- Smilex
- sadam36 
- LarryRotgut
- cerebrux 
- AllenWalkerHU 
- madrang

karolherbst:
@LarryRotgut @cerebrux @AllenWalkerHU @naryl @Telemaniaka @madrang and @ssokolow 
some of the names weren't linked above

Jookia:
I think due to the sheer amount of work they've done on it (99% or so) they should be allowed to dual license.

But can we justify a dual license? Why is the GPLv3 not good enough for Desura Ltd? Could this lead to in-house development being forked back?

karolherbst:
a BSD license would allow desura to not publish changes to desurium with GPL they have to.

Jookia:
Well, yeah. But why are we open source then?

karolherbst:
To help developing. But you are right, I don't see any good reason why we should not stay GPLv3 only.

ssokolow:
I could imagine a case for licensing under the LGPL or even the MPL to gain compatibility with other licenses, but I can't see any justifiable case for removing the copyleft protection on the existing code.

If nothing else, dual-licensing to BSD would be a horrible PR move as various people would see it as a betrayal and, when it comes to the human psyche, there's a reason Dante's Inferno reserved the lowest circle of hell for betrayers.

jcnix:
At least for my part, it doesn't matter to me.  You can make my patches public domain or proprietary for all I care.

TheSiege:
I do agree with @ssokolow LGPL should be good enough, from what I know we are talking about some updater code and that should then be covered

[...]

PR #601, opened 2013-09-11
add copyright notes to changed source files

### LINKED ISSUE #543: adding copyright information
I want to add copyright information in every source file.

I've created a branch for this here: https://github.com/karolherbst/Desurium/tree/copyright

But for me there are some questions:
- Should I add the real name/email of the developers as long as the name/email is shown on the github page?
- I prefer the format: Copyright (C) $years $name (<$email>) ($company). Does everybody aggree with that? stuff within () is optional. I've choosen this, because I don't have to touch lodles entry this way.

I will ask each for which I've added a copyright entry if he is aggreeing with what I've done.

karolherbst:
fixes #543

I wrote a script, which is handles all html, js, css, h, C++ and C files.

RC files were generated by msvc, but maybe we should clean them up and add the rc pattern to this script, too.

This script also added newlines at the end of every file without an empty new line. On Linux we should have empty lines at the end anyway.

### DIFF
--- a/build_out/data/bindings/host_binding.js
+++ b/build_out/data/bindings/host_binding.js
@@ -1,6 +1,6 @@
 /*
 Desura is the leading indie game distribution platform
-Copyright (C) 2011 Mark Chandler (Desura Net Pty Ltd)
+Copyright (C) Desura Ltd. <support@desura.com>
 
 This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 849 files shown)


Jookia:
We should wait on the copyright discussion before doing this.

karolherbst:
copyright is complete unrelated to license, because you are dealing with national law here. It doesn't matter what linden or desura decides I will have still the copyright on the changes, even after we would change the license to some propritary stuff.

Jookia:
Apologies, that was a typo. See #509 for the discussion.

Cheeseness:
For accuracy, I feel it's worth noting that that discussion is no longer current (and since it presupposed a dual licencing outcome when decisions hadn't been made, whether it accurately represented the situation is debatable).

After thinking about this, I've changed my position - maybe we should get copyright information into all relevant source files so that if the outcome of the upcoming more open licencing discussion requires some level of attribution, there'll be a clear indication of who's worked on what.

PR #606, opened 2013-09-16
Added virtual solution folders for visual studio

lodle:


Jookia:
Looks good to me. What about your @karolherbst ?

karolherbst:
Nothing of this should effect a non MSVC build at all, so I am fine with this.

The v8 stuff is just fixing a little compile error, when you have parallel builds (with jom instead of nmake for example)

Jookia:
Merging then.

Jookia:
If I can click merge properly, that is. :P

Issue #613, opened 2013-09-20
Clicking avatar does nothing


lodle:
Should show users profile

lodle:
Fixed with the removal of custom windows frame #658

PR #614, opened 2013-09-20
Fixed avatar not being clickable #613

lodle:


Jookia:
This causes a build error in Linux. See Travis CI. Otherwise it looks okay.

lodle:
God i hate gcc and templates ill get it fixed

Jookia:
Had to revert, broke tabs.

PR #616, opened 2013-09-20
Fixed issue with desura corrupting downloads #615

lodle:


Jookia:
@karolherbst This breaks on GCC (Travis CI at least), due to the override keyword. Are some flags not set?

karolherbst:
also you should try to reorder the commits with git rebase. So first this #define override thing, then the actual fixes. This will prevent some issues while rebasing bugs. We should try to keep every commit compileable on every platform and the fix for this isn't that hard.

karolherbst:
We should wait with this commit until gtest got merged

karolherbst:
cleaned up branch: https://github.com/karolherbst/Desurium/commits/bugfix615

karolherbst:
@lodle what do you think? https://github.com/karolherbst/Desurium/commit/cbecddf8b3fe903de00ed8ae2efbab54e8d76fcc

lodle:
Looks good karolherbst. Merge when ready

Issue #255, opened 2012-06-26
Make it to Debian/Ubuntu main repositories.


makson96:
This is the best way to make Desurium more popular among Linux users.

From Ubuntu side issue is here: https://bugs.launchpad.net/ubuntu/+bug/898357

From Debian side issue is here: http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=656903

As mentioned in both threads the main obstacle for this is to make Desurium use mainstream libraries instead of it's own patched ones. This is related to issue #115 .

stragu:
This would be very appreciated indeed. Right now, I have an issue with the Desura Linux client that makes it impossible to use it on Ubuntu, but Desurium works perfectly. It would be an awesome addition to the Ubuntu Software Centre, and making it before Steam arrives would be a good thing too, wouldn't it? ;)

karolherbst:
I will write a shell script to create a deb package without any problems

makson96:
Keep in mind, that Debian and Ubuntu build servers will not allow for net connection during compilation, so every time there is 'wget' command it will make compilation fail.

If you want I can describe, how I handle this in my ppa:
https://launchpad.net/~makson96/+archive/desurium

karolherbst:
we have flags to use locally stored archives

makson96:
So, has anyone contacted MOTU as suggested:
https://bugs.launchpad.net/ubuntu/+bug/898357/comments/19

One could also start thread about it on debian-devel-games mailing list:
http://lists.debian.org/debian-devel-games/

karolherbst:
steam got into jessie debian repository: http://packages.debian.org/jessie/steam

Issue #199, opened 2012-04-27
get Desurium stable without ares in curl


karolherbst:
I put this on wishlist.
Maybe this is possible, maybe not. The biggest advantage would be, that we don't have to build curl on our own and use the system one instead.

See #189

yurikoles:
Vote on this.

makson96:
I will give it a try during the weekend.
I will recompile software with: -DDEBUG=ON and -DWITH_ARES=OFF and test it on my Xubuntu 12.04 64bit. If I manage to crash the software, I will paste here terminal output.

makson96:
Hmm... So far I am unable to reproduce any issue. Everything is working just fine. Does anyone know what to do to break software with system curl (without c-ares)?

makson96:
OK. It breaks during game installation:

```
*** longjmp causes uninitialized stack frame ***: /home/makson/Dokumenty/Ubuntu_ppa/desurium/github/Desurium/install/lib/desura terminated
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(__fortify_fail+0x37)[0x7fee04fa2817]
/lib/x86_64-linux-gnu/libc.so.6(+0x10a78d)[0x7fee04fa278d]
/lib/x86_64-linux-gnu/libc.so.6(__longjmp_chk+0x33)[0x7fee04fa26f3]
/usr/lib/x86_64-linux-gnu/libcurl.so.4(+0xdff5)[0x7fee0118dff5]
/lib/x86_64-linux-gnu/libc.so.6(+0x364a0)[0x7fee04ece4a0]
/lib/x86_64-linux-gnu/libc.so.6(recv+0x6c)[0x7fee04f8cc6c]
/usr/lib/x86_64-linux-gnu/libxcb.so.1(+0xb30f)[0x7fedf9ffb30f]
/usr/lib/x86_64-linux-gnu/libxcb.so.1(+0xb826)[0x7fedf9ffb826]
/usr/lib/x86_64-linux-gnu/libX11.so.6(+0x3d499)[0x7fedfed75499]
/usr/lib/x86_64-linux-gnu/libX11.so.6(+0x3d561)[0x7fedfed75561]
/usr/lib/x86_64-linux-gnu/libX11.so.6(_XEventsQueued+0x5d)[0x7fedfed75cbd]
/usr/lib/x86_64-linux-gnu/libX11.so.6(XPending+0x5d)[0x7fedfed6735d]
[...]

PR #617, opened 2013-09-23
Bug 199

makson96:
This is tiny patch but with the huge impact. After applying it I have heavily tested Desurium with system curl on Ubuntu 12.04 (curl 7.22) and I am no longer able to reproduce issue #199 . After this I have set system curl as a default, what is big step to accomplish #255 .


karolherbst:
Thanks for digging into this and find a solution for the problem. I hope this will fix it for all Linux users. I found something, which should be in a third commit though, but this is just something little small.

karolherbst:
You don't have to follow my comments. But I think, that commits with information about where the changes are made will produce a much better git history.

Issue #621, opened 2013-09-26
remove arma hacks that no longer work due to open source


lodle:

karolherbst:
vote

Jookia:
Fixed.

PR #622, opened 2013-09-26
Removed arma hacks #621

lodle:


karolherbst:
seems nice. The Copyright stuff is unrelated to the pull request though. Maybe we should clarify how we should update the copyright? I could also move this change to my copyright branch, so we can collect everything related to it.

Jookia:
Has ARMA stuff been moved serverside? Also, could you split this in to two pull requests, one for copyright and one for ARMA hacks? Maybe look at adding it to #601?

lodle:
Gah, copyright stuff shouldnt of been done in this branch. :(

Issue #612, opened 2013-09-20
Ignore return code of tools


lodle:
There seems to be problem at the moment, where if a tool has an empty "return code" that it expects "0"

this is a BIG problem if it is the case, as for silent tools we kind of need to always ignore the return code, otherwise if something goes wrong there is virtually nothing the user can do (not sure what we can do about this?)

lodle:
Merged in https://github.com/desura/Desurium/commit/d7ad3b2092bac864e95de56acdf356e9f57af998#diff-6a95f0018873a25a91ec8d836396bcba and https://github.com/desura/Desurium/commit/c8cc3fbe1ee6630b6474773d9a8bf03e10b19e3a#diff-6a95f0018873a25a91ec8d836396bcba

PR #623, opened 2013-09-27
Fixed toolinfo to process result check properly #612

lodle:


Jookia:
Had to revert as it broke the build.

PR #627, opened 2013-09-28
a lot of cmake cleanups and windows stuff

karolherbst:
This pull request will clean up some cmake files and change some special windows behaviour:
1. instead of looking for *.cpp in cmake files we should list them. Adding new files won't trigger a cmake change, but with the second approach we know that we have to add them to the list
2. .gitignore updates
3. adding cmake functions for target installation. I added this, so we have the install logic for targets in one central cmake file (macros.cmake). Also these functions add some IDE support like moving target files and Categories
4. it reverts the dll and data paths to "bin" and "data" on windows
5. bumps the embedded boost version to 1.53, because msvc 2013 needs at least 1.52.
6. improves the patch.bat file for windows (detection of already applied patches
   I merged this branch together with my lodle_working branch:
7. add workaround for multiple boost installation on windows
8. gtest
9. compile some targets on windows with a static runtime
10. add updater code
11. removed mfc

Parts of it are done by lodle, and some of his parts were reworked a lot by me. The other cleaned up commits written by lodle will come over the next week.

I've tested it on Linux and windows (nmake and msvc).


karolherbst:
I will add some stuff to this pull request later, so please don't merge it yet

lodle:
Need this changset to build on windows https://github.com/lodle/Desurium/commit/c0d3d5b2af4ab48c7858605fb6d8bbd0a943abfa
https://github.com/lodle/Desurium/commit/e43382a5cd74704577e837f7d6ecb2c3d36779c1

karolherbst:
do you want to try out boost-1.54 and see if your issue boost issue is going away?

karolherbst:
Additional Info:

I know that there are no problems with boost and wxWidgets on msvc10 and msvc11. Because msvc12 is still in beta, we should check if newer versions of both libraries are fixing the errors. I know that wxWidgets complained about not knowing msvc11, but this did not result into errors.

karolherbst:
I read, that it is possible to install different msvc installation side by side as long as you install the oldest one first. I think I will install msvc12 here and test is myself.

lodle:
Yes you can install them side by side and depending on what visual studio command line you use to run cmake it will select a different visual studio version. 

The complaints are fine but we need to tell boost what version to compile with.

karolherbst:
Are there any specific reason why we have to tell boost even for msvc11 and msvc10?

lodle:
bjam doesnt pick it up.

karolherbst:
I've reported the boost bug: https://svn.boost.org/trac/boost/ticket/9214

lodle:
Missing dumpgen.exe in root folder (neede for official builds)
Missing extra bin pack  SetupDownloadUrl(DESURA_EXTRA_BIN "desura_extra_bin.zip" fdb2a4dd704a90e5174c64d3eaf8392b) 
browser_loading_spinner.gif is missing

karolherbst:
the browser_loading_spinner.gif will be added with my cmake_install_rework request after this one. I will take care about the other issues here

karolherbst:
@lodle: one thing, dumpgen needs to be ported away from MFC

karolherbst:
this should be ready now

PR #629, opened 2013-09-30
Added option to disable news popup #628

ghost:


src/shared/uicore/code/MainApp_internal.cpp
```diff
@@ -216,6 +216,11 @@ void MainApp::showNews()
 		return;
 	}
 
+	if (! gc_enable_news_popups.getBool())
```
lodle: Think you need to still clear news items and gift items (see code below) as if you dont and a user enables this they might get a lot of news items.

lodle: Also want gift updates to still make it

PR #631, opened 2013-10-02
Updated release build to build with debugging information #630

lodle:


karolherbst:
We should rework the DEBUG option anyway, I just have to figure out how to deal with external projects on windows with the cmake internal build types

lodle:
This is for release builds though. Need the pdb files to debug crash dumps

karolherbst:
Yeah, I know. This path can be used for creating releases, but I would like to have a more cmake aware solution. There should be a way of telling cmake when do build External builds in which way without having too much pain.

karolherbst:
any updates here?

lodle:
Im not sure what you want. This is the cmake solution for this.

karolherbst:
I wanted to know what you ideas are about using RelWithDebInfo only for official builds

lodle:
Oh can do that. Is there an issue with having it for non official builds? Debug info is always useful

karolherbst:
yeah, but distributions want to set the flags on their own anyway

PR #634, opened 2013-10-02
Fixed boost throwing an exception parsing date string

lodle:


karolherbst:
I am fine with this, but we should try to be more strict about proper commit messages. There are some changes wich belongs in the first commit, but are in the second one.

We should talk about QA and what is required before commits get merged.

lodle:
Looking back the first commit is not needed at all. Im open to ideas but it doesnt bother me that much that a bit of noise gets into the commits.

karolherbst:
yeah, that's the reason why we should talk about it, so we have an objective set of rules we can al stick to

karolherbst:
Is there a way how to reproduce this bug, so I can check this?

lodle:
Look at the unit tests?

karolherbst:
yeah, :(

lodle:
With out this fix try installing a game. It should fail on setting the date stamp on the file every time and spam the console.

karolherbst:
ahh, I actually saw this I think

lodle:
And you sat there and did nothing :D

karolherbst:
This was some days ago, actually after this request :D

karolherbst:
@lodle cleaned up branch is here: https://github.com/karolherbst/Desurium/tree/bugfix633

karolherbst:
merged in de8644fa53a7998dcc7f0a569e225dec2c39abab

Issue #639, opened 2013-10-07
boost causes desura to crash due to boost bug 6320


lodle:

karolherbst:
fixed since 03d9592b579552281af3a4d83b34b298d100f77f

PR #641, opened 2013-10-07
Fixed boost causing desura to crash due to boost bug 6320 #639

lodle:


karolherbst:
should we try to use the dll version of boost instead? We could check if this is increasing or decreasing the package size of desurium.

lodle:
we should as that is what we used in desura before. For now this will fix it.

PR #642, opened 2013-10-07
Fixed black lines showing on breadcrumbs due to expand taking extra space #213

lodle:


karolherbst:
this is a fix to #213, isn't it?

karolherbst:
I've tested it on windows, seems like it fixed it. Nice work. I will test it on Linux later.

Issue #373, opened 2012-12-10
Text in error dialog boxes should be selectable for copy-paste


ssokolow:
Whenever I encounter an error dialog in Desura (for example, when trying to download Bean's Quest via the client), I have to manually transcribe the error.

That leaves a LOT of room for people to mistype numeric IDs which could be essentially for quickly resolving problems.

I'm not sure how they do it but a lot of GTK+ applications use label widgets which support being selected and copied to the clipboard and it'd probably be a good idea for Desura to do the same with its dialogs when using wxGTK as its backend.

(For example, every single text widget in the About dialogs for [GNOME Character Map](https://live.gnome.org/Gucharmap), [Leafpad](http://tarot.freeshell.org/leafpad/), [LXTerminal](http://wiki.lxde.org/en/LXTerminal), [PCManFM](http://wiki.lxde.org/en/PCManFM), and [Comix](http://comix.sourceforge.net/). I suspect it may be a GTK+ common dialog like Open/Save)

PR #643, opened 2013-10-07
Added control + c (copy) to message dialogs to copy text #373

lodle:


karolherbst:
Shouldn't that be handled automatically? I am worried, that we are doing so much on our own and force specific key combinations, which seems to be standard, but specially on Linux you are able to rebind nrearly everything.

Isn't there a way to add the copy shortcut without knowing which it is?

lodle:
this is only for error dialog's and control + c is very well known. Im not to worried about custom bindings.

karolherbst:
ask the guys who are using dvorak ;) I just would loke to know if there is any other possibility as checking against WXK_CONTROL_C. If there is no, I will be fine with that.

lodle:
From a quick google, no

Jookia:
Maybe we should open a new issue for compatibility with keyboard shortcuts and push that back to another time.

karolherbst:
I will test this later

karolherbst:
cleaned up branch: https://github.com/karolherbst/Desurium/commits/bugfix373

lodle:
moved to https://github.com/desura/Desurium/pull/682

Issue #640, opened 2013-10-07
Desura fails to save settings on login (like auto login)


lodle:

karolherbst:
works for me on windows an linux without the commit

karolherbst:
fixed since 689cf5a9a31657556c18445664f858f95982abcb

PR #644, opened 2013-10-07
Forced desura to save settings on login to ensure auto login settings al...

lodle:
...ways apply #640


karolherbst:
Seems save to me

karolherbst:
what happens with a wrong password?

lodle:
this is for the auto login settings, as if the app doesnt close nicely the settings dont get saved at all causing auto login to be disabled next startup

karolherbst:
I am just worried what happens if you check auto login, fill in username and a wrong password and quit the client after that

lodle:
It will just return to the login form. Not something im worried about. Auto login requires one good login to work

karolherbst:
okay

PR #646, opened 2013-10-08
Added support for running with debuggers easily.

Jookia:


cmake/scripts/run.sh
```diff
@@ -12,11 +12,11 @@ if [[ "@RUNTIME_LIB_INSOURCE_DIR@" = /* ]]; then
 	export LD_LIBRARY_PATH="@RUNTIME_LIB_INSOURCE_DIR@/"
 	# also add our runtime lib directory to path
 	export PATH="$LD_LIBRARY_PATH:$PATH"
-	@RUNTIME_LIB_INSOURCE_DIR@/@CURRENT_TARGET@ $@
+	$DEBUGGERS @RUNTIME_LIB_INSOURCE_DIR@/@CURRENT_TARGET@ $@
```
karolherbst: I would rename it to DESURA_DEBUGGER. We could do some more stuff with it later this way. Maybe we want to attach a debugger on a desurium crash at runtime later. And setting something like DEBUGGERS is too general

Jookia: It's meant to be general so it can encompass things like Valgrind, GDB, etc. As for DESURA_DEUBBGER? Maybe. I'm a bit skeptical seeing as I don't think people are going to be defining DEBUGGERS in their bashrc, so I don't there's danger of naming conflicts.

karolherbst: but still. If other projects would think the same way, maybe there would be a reason to set DEBUGGERS in bashrc

karolherbst:
Am I the only one with the feeling we should take another variable name?

lodle:
Desura prefix is good

Jookia:
Added.

Issue #288, opened 2012-08-06
some GUI bugs with wxWidgets-2.9.4


karolherbst:
with wxWidgets-2.9.4 there are some displaying issues with CEF. Changing size of windows let the internel CEF view disappear and witching between the content views will bring them back after the second switch.

task list
- [x] resizing bug
- [ ] crash after game update

TheSiege:
also for me there are a lot of crashes when verifying; or installing games in desurium

Mailaender:
wxWidgets 2.9.X compatibility is also important for inclusion into openSUSE games: https://build.opensuse.org/request/show/136219

TheSiege:
please note that we had no choice about wxwidgets 2.9; that design decision was made before opensourcing of desura; so we were not neccessarily aware that this kind of breakage occurs

karolherbst:
I think, this bug should wait until wxGTK 3.0 is released.

karolherbst:
We should bundle wxGTK, because it will cause less damage in repositories, too. See #423

makson96:
Has anyone tried wxWidgets 2.9.5 yet?

karolherbst:
I did: https://github.com/karolherbst/Desurium/commit/eed1878709614a2089bc9e8d524e9bb5b88b0a4a

makson96:
Does it still has GUI bugs?
Do you want me to test it during the weekend?

karolherbst:
every test is welcome. I think there are some pixel errors around buttons and resizing is very slow

makson96:
Here is documentation of the issues with wxWidgets 2.9.5:

1.
http://img534.imageshack.us/img534/3562/de9b.png
Ping dots in the corners of the buttons.

2.
http://youtu.be/adiLbKpUPhc
Resize update slowly.

I have not found any other issues. I will continue testing.

P.S. I wonder how many bugs could we close if someone capable would update CEF.

TheSiege:
the slow resize looks to me like the one I already have (with other 2.9) ... I will test that in a bit

karolherbst:
we have this issue since 2.9.4. We should check if we are able to install games

makson96:
Installing, running and uninstalling games works fine (tested on World of Goo). Updating games also works fine for me (tested on Beep).

karolherbst:
okay, this sounds better. Thanks for testing.

hasufell:
seems to work

the only regression I see are those red small dots

[...]

PR #652, opened 2013-10-14
Use Wx3.0 instead of bundled 2.9.3 if present in the system.

makson96:
wxWidgets developers manage to point out fixes for the breaks in compatibility with wxWidgets 2.9.5.
http://trac.wxwidgets.org/ticket/15529

With this pull request if wxWidgets 3.0 is present in the system, than it will be used instead of bundled 2.9.3. On Linux it works great on wxWidgets 3.0.0-r1 without pink dots regression: #288 and still provide backward compatibility with wxWidgets 2.9.3.

I publish this pull request for testing purposes. I hope that when wxWidgets 3.0 comes out it will be working great with this changes and then we can merge it to the master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
   - mkdir build
   - cd build
 script:
-  - cmake .. -DBUILD_CEF=off -DWITH_ARES=off -DDEBUG=on -DFORCE_BUNDLED_WXGTK=off
+  - cmake .. -DBUILD_CEF=off -DWITH_ARES=off -DDEBUG=on
   - make -j4
   - ctest --output-on-failure .
 notifications:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@
 #   CURL_URL                            URL to curl archive
 #   DEPOT_TOOLS_URL                     URL to depot_tools archive
 #   V8_URL                              URL to v8 archive
-#   WXWIDGET_URL                        URL to wxWidget archive, has to be 2.9.3
+#   WXWIDGET_URL                        URL to wxWidget archive, has to be 3.0
 #
 # cef options:
 #   H264_SUPPORT    OFF                 build ffmpeg with mpeg-4 support. Be aware of patent and license problems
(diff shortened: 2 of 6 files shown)


makson96:
Well I was thinking, that we will change bundled wxWidgets after 3.0 final. If you want me to switch bundled version now, could you upload archive with 3.0.0-r1 to our sourceforge page and give me url and md5.
I can also remove compatibility with 2.9.3 and 2.9.4 if you want.

karolherbst:
if it works with 3.0.0-rc1 then you can use this link:
http://garr.dl.sourceforge.net/project/wxwindows/3.0.0-rc1/wxWidgets-3.0.0-rc1.tar.bz2

the checksum will be calculated by cmake and printed out because of a mismatch

Lodle:
Please use this url: https://s3-us-west-2.amazonaws.com/lecs.desura.lindenlab.com/wxWidgets-3.0.0-rc1.tar.bz2
Md5: a4f74ef439b7a95d76c699b3a198097c

karolherbst:
also some versions have to be changes in this file: https://github.com/desura/Desurium/blob/master/cmake/modules/BuildwxWidgets.cmake

karolherbst:
But I would do that if you want

makson96:
If you are not in the hurry with this you can focus on more complicated tasks :)

I can do it later today. 2 questions:
1. After bundling 3.0.0-r1 do you want me to remove hacks, which provide compatibility with 2.9.3 (gxImageControl, UICoreMain)?
2. "find_package(wxWidgets 3.0 COMPONENTS richtext ${REQUIRED_IF_OPTION})" always return TRUE, no matter of wxWidgets version. How do you propose to fix this?

karolherbst:
1. yes, we want to target 3.0.0 final and distros doesn't want to use older versions, too. I don't see a reason why we should keep that for compatibility reasons.
2. this is a cmake problem and should be reported there.

karolherbst:
here I wrote a patch for the 3.0 transition: https://github.com/karolherbst/Desurium/commit/15634cd7c172a64b5343f4559d26e11485b4d50e feel free to pick it

lodle:
Has some one tested this on windows yet?

makson96:
Nope.
Beside Windows, wx3.0 is now working also against Gtk+ 3, so we should test this platform too. I only tested it for Linux and Gtk+ 2 yet.

karolherbst:
I am sure this will work on windows, but I will test it when I have some time

makson96:
I have removed SetLocale. If anybody will find bug concerning this, we can think of bringing it back properly.

karolherbst:
I think I will cleanup stuff an create a new pull request with some additional changes (I will keep your author tags)

karolherbst:
new request #653

PR #653, opened 2013-10-14
Use Wx3.0 instead of bundled 2.9.3 if present in the system.

karolherbst:
original request by @makson96 #652

wxWidgets developers manage to point out fixes for the breaks in compatibility with wxWidgets 2.9.5.
http://trac.wxwidgets.org/ticket/15529

With this pull request if wxWidgets 3.0 is present in the system, than it will be used instead of bundled 2.9.3. On Linux it works great on wxWidgets 3.0.0-r1 without pink dots regression: #288 and still provide backward compatibility with wxWidgets 2.9.3.

I publish this pull request for testing purposes. I hope that when wxWidgets 3.0 comes out it will be working great with this changes and then we can merge it to the master.


Jookia:
Works well for me.

karolherbst:
did you also notice compile time improvements with embedded wxgtk?

karolherbst:
I have removed the dependency for richtext on system wxWidgets. Is anybody able to check if desurium compiles with a system installed wxWidgets-2.9.5+ ?

karolherbst:
I also brought back compatibility with older wxWidgets then 2.9.5, because some package repositories are still at 2.9.4 max and we do't want to break compilation there.

makson96:
I will check this on system wxWidgets 3.0.0-r1 (Gtk+2) later today.

karolherbst:
@makson96 any updates?

makson96:
Well, I did not manage to test it yesterday. I will do that, when I will have some free time to play with Desurium.

karolherbst:
okay, thanks

karolherbst:
for reference: http://www.cmake.org/Bug/view.php?id=10694

makson96:
This branch works without any noticeable regression on system installed wxWidgets 3.0.0-r1.

karolherbst:
I will test this on windows

karolherbst:
windows doesn't seems to work with it

makson96:
Could you please move now to wx 3.0-rc2?

makson96:
We should update now to wx 3.0 final.

lodle:
ill give this a go start of next year

Issue #658, opened 2013-10-15
Remove custom windows chrome causing many issues


lodle:

karolherbst:
Wich issues for example? I only noticed, that the windows looking very windows 7 like on a windows 8 machine, but nothing else.

lodle:
We had around 10 issues on our internal bug tracker with quirks that windows did with normal windows and the desura windows didnt follow like drag to top of screen to fully maximize.

karolherbst:
okay

PR #659, opened 2013-10-15
Remove custom windows chrome causing many issues #658

lodle:


karolherbst:
I like cleaning up, but what are the effects and what will be different?

src/include/wx_controls/gcCustomFrame.h
```diff
@@ -68,154 +66,19 @@ class gcCustomFrame : public T, public FrameIcon
...
-			m_pCustomFrame->destroy();
-			m_pCustomFrame->onActiveEvent -= delegate(&onActiveEvent);
-		}
-#endif
-
-		return T::Destroy();
-	}
-
-	void init(gcCustomFrameImpl* custFrame)
+	virtual ~gcCustomFrame()
```
karolherbst: do we still need this destructor? I think it would be safe to remove it fully

lodle:
It will look like a normal windows application instead of having the integrated title bar into the top of the app

karolherbst:
I've tested it on windows 8, it seems to work fine.

PR #666, opened 2013-10-16
Tinyxml2::LoadFile returns 0 for success instead of true thus updater is broken #665

lodle:


lodle:
https://github.com/lodle/Desurium/commit/8057f3f6b593a40f1ead5716cde4d25c5077972e

karolherbst:
okay

PR #668, opened 2013-10-17
needed changes for coveralls.io support and enable it in .travis.yml

karolherbst:
just see for yourself: https://coveralls.io/r/karolherbst/Desurium


karolherbst:
there is an error somewhere

PR #669, opened 2013-10-17
big rework of the cmake install command and move the build_out directory

karolherbst:


Jookia:
@lodle What do you think of this?

lodle:
Looks good, needs to be rebased against latest so it can be auto merged

PR #674, opened 2013-10-18
Changed urls to use s3

lodle:


karolherbst:
again, I don't see any benefits here.

lodle:
its useful for linden teamcity builds as it cuts a lot of time off.

PR #675, opened 2013-10-18
Converted boost tests into gtests #647

lodle:


src/shared/uicore/code/UICoreEntry.cpp
```diff
@@ -94,6 +96,16 @@ class UICore : public UICoreI
 
 		m_bExitCodeSet = false;
 		m_iExitCode = 0;
+
+#ifdef DEBUG
+#ifdef WITH_GTEST
+#ifdef WIN32
```
karolherbst, line 102: why not in one line?
#if defined(DEBUG) && defined(WITH_GTEST) && defined(WIN32)

src/shared/unittests/code/util/MD5_test.cpp
```diff
@@ -0,0 +1,28 @@
+#define BOOST_TEST_MODULE MD5_test
```
karolherbst: can be removed

src/shared/unittests/code/util/gcBuff_test.cpp
```diff
@@ -0,0 +1,62 @@
+#define BOOST_TEST_MODULE gcBuff_test
```
karolherbst: can be removed

Jookia:
I think this breaks CMake's amount of tests being run.

karolherbst:
yeah, only the gtest test, but this is okay. On error the output gtest output will be printed on travis-ci

PR #677, opened 2013-10-18
Added precompiled headers for windows build #676

lodle:


Jookia:
@karolherbst What do you think of porting the CMake code to Linux? Currently CMake doesn't have cross-compiler PCH support but it'd be useful to try and set it up.

lodle:
I think its worth creating a separate item for it and a linux guru can then do it.

Jookia:
Yeah, it has an if(MSVC) tag so it's close enough to abstracted for me.

PR #679, opened 2013-10-18
Fixed crash due to invaild ico file and games for windows #632

lodle:


Jookia:
Can't test it on Windows, but it compiles on Travis CI and the code is trivial. Merging.

PR #681, opened 2013-10-18
Added alternative login providers for desura (steam, google, twitter, facebook) **Requires desura web update to work**

lodle:


karolherbst:
the first commit is rather big and can be hardly reviewed. Would it be possible to split it in smaller commits, so it we are able to find bugs through git bisect easier in the future?

PR #682, opened 2013-10-18
Added control + c (copy) to message dialogs to copy text #373

lodle:


Jookia:
Could we perhaps have this not as WXK_CONTROL_C but something like keys::copyKey() ? I'm just worried about in future if we have to abide to UX guidelines.

lodle:
What is keys::copyKey()? i dont see why this is such a big issue. If it needs to change we can always revisit it later.

Jookia:
It's a hypothetical thing. Are we sure in the future we won't be adding more keybindings?

lodle:
If we did i think we would look into a solution that would be more generic than hard coded values. For now this is fine

PR #714, opened 2014-01-08
Fix fork license to LGPL

makson96:
lodle make a typo. New fork license is LGPL, not GPL. This might be very misleading, so should be fixed.


lodle:
Thanks

